PR #17, opened 2019-08-27
Featue/16 adding members to achievements

### LINKED ISSUE #16: Adding the ability to add members to achievements.
Adding:
- [x] Ability to add members to achievements
- [x] Ability to see members added to an achievement
- [x] Ability to see achievements per category

HansNorg:
Closes #16


achievements/serializers.py
```diff
@@ -7,10 +7,10 @@ class AchievementSerializer(serializers.ModelSerializer):
     class Meta:
         model = Achievement
         fields = "__all__"
-        depth = 0
```
helaan: Question: what does the depth do?

achievements/migrations/0002_auto_20190826_2224.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.3 on 2019-08-26 20:24
```
helaan: So many migrations, isn't there a better way?

helaan:
Please rebase and fix CI build

achievements/migrations/0002_auto_20190826_2224.py
EricTRL: You can squash the migrations (https://docs.djangoproject.com/en/2.2/topics/migrations/#migration-squashing). Might be good to keep the initial migration (the one regarding the membership file) as a separate migration though.

achievements/serializers.py
EricTRL: It's used for foreign keys/many-to-many relations. When the depth is 1, instead of obtaining the foreign key when requesting the object, the associated object is obtained instead.
If the depth is 2, then associated objects of associated objects are loaded in, etc. (Not sure how this interacts with many-to-many relations though)

HansNorg: It makes sure that it is possible to see the name and description of the associated category. With a depth of 0 you can only get the id of that category.

helaan: That is an interesting way to solve the N+1 query problem: instead of lazy-loading items, always load them or never load them depending on the model configuration.

achievements/migrations/0002_auto_20190826_2224.py
HansNorg: It might be an idea to just not worry about it now and when we are ready to release the first version of the app we just squash them? This way we will not lose a lot of time on squashing and re-squashing again and again?

helaan: If you look at your CI build out, you can see that it is related to database migrations. Therefore I suggest you fix this issue while you are fixing the build.

Also, ever heard of the Broken Window theory?

HansNorg: No and I also do not know what I would have to do to 'fix the build'

helaan: Looking at the build output, which you can find in the check menu, you can see that the build crashes with `django.db.utils.OperationalError: no such table: achievements_test`. No clue what happened there.

HansNorg: I will have a look then

achievements/tests.py
```diff
@@ -1,3 +1,75 @@
-from django.test import TestCase
+from django.test import TestCase, Client
+from .serializers import AchievementSerializer, CategorySerializer
+from .models import Achievement, Category
+from membership_file.models import Member
+from membership_file.serializers import MemberSerializer
 
 # Create your tests here.
+class TestCaseAchievementFrontEndViews(TestCase):
+    @classmethod
```
EricTRL: should remove `@classMethod` (it's only used for `setUpTestData(..)`)

achievements/tests.py
```diff
@@ -1,3 +1,75 @@
...
+            "house_number": "5",
+            "city": "Eindhoven",
+            "country": "The Netherlands",
+            "postal_code": "5612 AH",
+            "member_since": "1970-01-01",
+        }
+        self.member = Member.objects.create(**self.memberData)
+        Member.save(self.member)
+
+    def test_case_view_all_categories(self):
```
EricTRL: Missing a description of what the testcase tests. E.g. `#Tests if the page that shows all the achievement-categories can be accessed`

achievements/tests.py
```diff
@@ -1,3 +1,75 @@
...
+        response = self.client.get('/achievements/', data=testSerializer, secure=True)
+        self.assertEqual(response.status_code, 200)
+
+    def test_case_view_specific_achievement(self):
+        testSerializer = (AchievementSerializer(Achievement.objects.all().first())).data
+        response = self.client.get('/achievements/1', data=testSerializer, follow=True, secure=True)
+        self.assertEqual(response.status_code, 200)
+
+    def test_case_view_specific_member(self):
+        #testSerializer = (MemberSerializer(Member.objects.all().first())).data
```
EricTRL: commented out code

achievements/urls-frontend.py
```diff
@@ -6,6 +6,8 @@
     path('', views.viewAllAchievements, name='achievements-frontend'),
     path('<int:id>/', views.viewSpecificAchievement, name='achievement-frontend'),
 
-    path('categories', views.viewAllCategories, name='categories-frontend'),
+    path('categories/', views.viewAllCategories, name='categories-frontend'),
     path('categories/<int:id>/', views.viewSpecificCategory, name='category-frontend'),
+
+    path('members/<int:id>/', views.viewSpecificMember, name='achievement-member-frontend'),
```
EricTRL: Doing this allows everyone to see the achievements of any member (as anyone can access this page). I'm not sure if we want this.

HansNorg: This is indeed something we need to think about, but at the moment people are able to see all people that have gained an achievement via the view_all_achievement page already.

PR #25, opened 2019-09-13
Feature/5 login logout register

### LINKED ISSUE #5: Login, Logout, and Register
Allows users to login, logout, and register new accounts.
Board Members being able to link user accounts to information in the membership file is NOT yet included here.

EricTRL:
Closes #5


core/forms.py
```diff
@@ -0,0 +1,70 @@
...
+# and allows Inactive users to login
+class LoginForm(AuthenticationForm):
+
+    def clean(self):
+        # Obtain username and password
+        username = self.cleaned_data.get('username')
+        password = self.cleaned_data.get('password')
+
+        # Check if both are provided
+        if username is not None and password:
```
HansNorg: What should this if statement do? It looks like it will only be True if password is True?
Should it be if (username is not None) and (password is not None)?

core/templates/core/user_accounts/account.html
```diff
@@ -0,0 +1,16 @@
...
+
+{% block content %}
+    Hi {{ user.username }}!
+    This is your account page.
+
+    <br/>
+    This is your email: {{ user.email }}
+
+    <br/><br/>
+    Click <a href="{% url 'core/homepage' %}">here</a> to go back to the homepage.
```
HansNorg: Doesn't there need to be a <p> </p> or something similar around this text?

core/templates/core/user_accounts/password_change/password_change_done.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'core/base.html' %}
+{% load i18n static %} <!-- Serves translation -->
+
+{% block title %}
+  Squire - Password Change Done
+{% endblock title %}
+
+{% block content_title %}<h1>{{ title }}</h1>{% endblock %}
+{% block content %}
+<p>{% trans 'Your password was changed.' %}</p>
```
HansNorg: This is the only translation in the whole template, shouldn't everything be translated?

core/templates/core/user_accounts/password_change/password_change_form.html
```diff
@@ -0,0 +1,53 @@
+{% extends 'core/base.html' %}
+{% load i18n static %} <!-- Serves translation -->
+
+{% block title %}
+  Squire - Password Change
```
HansNorg: Everything else in this template is translated, isn't it more logical to translate the title as well?

core/forms.py
EricTRL: This statement utilises python's type conversion. E.g. Expressions such as `False`, `""` (the empty string), `None`, or `[]` (the empty list) will all result in `False`.
These should not be valid for a username either, so it should actually be `if username and password:`, good catch!

core/templates/core/user_accounts/account.html
EricTRL: All these templates are placeholders; they are ugly (I'd say :P) and are only there to showcase the functionality, which is why not much attention was given to this.

core/templates/core/user_accounts/password_change/password_change_done.html
EricTRL: See the comment at https://github.com/EricTRL/Squire/pull/25#discussion_r326639283

core/templates/core/user_accounts/password_change/password_change_form.html
EricTRL: See the comment at https://github.com/EricTRL/Squire/pull/25#discussion_r326639283

core/templates/core/user_accounts/account.html
HansNorg: Do they work without?

README.md
```diff
@@ -11,6 +11,7 @@ A re-introduction of the web application for ESRG Knights of the Kitchen Table.
 1. Install the dependencies: `pip install -r requirements/dev.txt`. These dependencies include common dependencies (such as *Django*) as well as dev-dependencies that speed up or ease the development process (such as *coverage.py*). For more information about dependencies, view the *Dependencies* section below.
 1. Setup the database by running `python manage.py migrate`. This ensures your database can store the items we expect to store in it.
 1. Start the server: `python manage.py runserver`. This starts a web server, which you can access using your webbrowser and going to `localhost:8000`.
+1. If wanting to use functionality that involves sending emails (such as resetting a password), then you'll also need to run the following command in another command prompt/terminal: `python -m smtpd -n -c DebuggingServer localhost:1025`. This will mimic an smtp email server. Any emails that would normally be sent will instead show up in this terminal.
```
HansNorg: I do not know if ```python -m smtpd -n -c DebuggingServer localhost:1025``` is correct

core/tests.py
```diff
@@ -0,0 +1,329 @@
...
+        # Ensure that only one error was given
+        self.assertTrue(form.has_error('password2'))
+        self.assertEqual(len(form.errors.as_data()['password2']), 1)
+   
+    # Test if a registering fails if username or email was already chosen by another user
+    def test_form_duplicate_field(self):
+        form_data = {
+            'username': 'schaduwbestuur',
+            'password1': 'bestaatookniet',
+            'password2': 'nomatch',
```
HansNorg: Shouldn't the passwords match for this test?

README.md
EricTRL: It was taken straight from the [documentation](https://docs.djangoproject.com/en/2.2/topics/email/#configuring-email-for-development).

core/templates/core/user_accounts/account.html
EricTRL: Assuming you mean that the template works without the `<p></p>`, yes. HTML does not care much about syntax and accepts pretty much anything. HTML-files can even consist of text only, though that's not really used outside of development/placeholders.

core/tests.py
EricTRL: Yes they should, nicely spotted!

PR #49, opened 2020-08-12
Feature/48 icalendar feed

EricTRL:
Admins can add (recurring) activities in the admin panel.
Anyone can view published activities on root/calendar. You can click on an event for its details (E.g. recurrence-information, location, and description.). The layout is similar to that of Google Calendar and is created by the FullCalendar javascript-library.
Anyone can obtain an .ical file via root/api/calendar/ical. This file can be imported in calendar software such as Google Calendar. When in production, people can add this URL itself to their software, resulting in an automatically-syncing calendar. (This cannot be done at the moment as you cannot import localhost as a URL unfortunately)

There is currently no way to modify (or cancel) a single occurrence of a recurring activity; I'm leaving this up for another PR as to not make this PR even larger.


.gitignore
```diff
@@ -122,6 +122,7 @@ db.sqlite3
 # Squire-specific gitignore (Secret Keys, etc.)
 /squire/secret_key.txt
 htdocs/media/images/achievements
+/components
```
helaan, line 125: Q: What's in the components folder that it should be ignored?

activity_calendar/feeds.py
```diff
@@ -0,0 +1,237 @@
...
+    vtimezone_daylight_tzname = "CEST"
+    vtimezone_daylight_dtstart = datetime(1970, 3, 29, 20)
+    vtimezone_daylight_rrule = build_rrule_from_text("FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU")
+
+    # Winter Time
+    vtimezone_standard_tzoffset_from = timedelta(hours=2)
+    vtimezone_standard_tzoffset_to = timedelta(hours=1)
+    vtimezone_standard_tzname = "CET"
+    vtimezone_standard_dtstart = datetime(1970, 10, 25, 3)
+    vtimezone_standard_rrule = build_rrule_from_text("FREQ=YEARLY;BYMONTH=10;BYDAY=-1SU")
```
helaan, line 115: Isn't there a nicer way to do this? Timezones are annoying, can't you refer to the Olsen Timezone Database to retrieve this data? Otherwise we'll need to patch this app when DST introduction is moved or abolished

activity_calendar/feeds.py
```diff
@@ -0,0 +1,237 @@
...
+                        datetime.combine(timezone.localtime(dt).date(),
+                        event_start_time)
+                    ),
+                    item.recurrences.exdates
+            ))
+            return item.recurrences.exdates
+
+    # RECURRENCE-ID
+    def item_recurrenceid(self, item):
+        return None
```
helaan, line 208: Looking at the monkey patch at line 15, what does this do and is it needed?

activity_calendar/static/fullcalendar/LICENSE.txt
```diff
@@ -0,0 +1,22 @@
+MIT License
```
helaan, line 1: Expat License is okay and as we don't have a full JS pipeline yet I think it is fine to vendor this library. 

Can you add a link to the upstream source of this library, which version you have imported and which changes you have made in these files?

.gitignore
EricTRL: A folder that was left behind as a result of experimentation with different libraries. Seems like I accidentally pushed the modified .gitignore!

activity_calendar/feeds.py
EricTRL: At the moment is is not. However, once we support occurrence-editing of a recurring event, then we need to be able to define a RECURRENCE-ID. I added this monkey-patch before I decided that I was splitting event-editing and event-viewing/ical into two separate PRs as to not make them insanely large.
I never bothered to remove it as it was going to be in the next PR anyways.

helaan (approved):

PR #51, opened 2020-08-29
Feature/50 subscriptions

### LINKED ISSUE #50: Activity Subscriptions
Users should be able to subscribe for activities.
Activities have a number of slots that users can subscribe to.
Users may be able to create slots as well.
There could be a limit on the number of slots, participants per slot, and number of participants in general.
There could also be a limit on the number of slots that a single user may subscribe to.

EricTRL:
Slots with different constraints can be created.
Users can register for slots.

Closes #50


activity_calendar/migrations/0002_auto_20200814_1431.py
```diff
@@ -1,88 +0,0 @@
-# Generated by Django 2.2.3 on 2020-08-14 12:31
```
helaan, line 1: Can we safely delete migrations?

activity_calendar/models.py
```diff
@@ -5,11 +5,19 @@
...
 from recurrence.fields import RecurrenceField
 
-from core.models import ExtendedUser as User
+from core.models import ExtendedUser as User, PresetImage
 
 # Models related to the Calendar-functionality of the application.
 # @since 29 JUN 2019
 
+# Not now, but a later time (used as a default value below)
+def later_rounded():
```
helaan, line 14: nit: why not implement using now_rounded?

activity_calendar/models.py
```diff
@@ -22,45 +30,232 @@ class Meta:
...
+    #                                          Until that time they do look like real slots though (in the UI)
+    # - Users: Slots can be created by users. Users can be the owner of at most max_slots_join_per_participant slots
+    SLOT_CREATION_OPTIONS = [
+        ("CREATION_NONE",   "Never/By Administrators"),
+        ("CREATION_AUTO",   "Automatically"),
+        ("CREATION_USER",   "By Users"),
+    ]
+
+    # The maximum number of dummy slots that show up if slot_creation=CREATION_AUTO
+    # Dummy slots do not exist in the DB itself, and match their title, description, etc. with the parent activity
```
helaan, line 79: Why do dummy slots not exist in the database?

activity_calendar/static/css/fullcalendar-customisation.css
```diff
@@ -42,3 +42,19 @@
     white-space: pre-line;
 }
 
+/*
+SVG Icon colours
+Generated using from hexcodes using: https://codepen.io/sosuke/pen/Pjoqqp
```
helaan, line 47: nit: which hexcodes were used to start with?

activity_calendar/forms.py
```diff
@@ -0,0 +1,38 @@
...
+class ActivitySlotForm(ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args,**kwargs)
+        self.fields['image'].queryset = PresetImage.objects.filter(selectable=True)
+        self.fields['image'].label_from_instance = lambda image: image.name
+
+    def is_valid(self):
+        ret = forms.Form.is_valid(self)
+
+        # Users can only create slots for at least 1 participan
```
helaan, line 25: t

activity_calendar/forms.py
```diff
@@ -0,0 +1,38 @@
...
+        # NB: At this point the data is a string
```
helaan, line 26: whyyyy

activity_calendar/migrations/0002_auto_20200814_1431.py
EricTRL: This one was only introduced in this branch, so in this case, yes!

helaan: I disagree. https://github.com/EricTRL/Squire/blob/master/activity_calendar/migrations/0002_auto_20200814_1431.py

However that commit has not yet been deployed so we're technically in the clear...

activity_calendar/models.py
```diff
@@ -22,45 +30,232 @@ class Meta:
...
+    MAX_NUM_AUTO_DUMMY_SLOTS = 1
+
+    # The way slots should be created
+    slot_creation = models.CharField(
+        max_length=15,
+        choices=SLOT_CREATION_OPTIONS,
+        default='CREATION_AUTO',
```
helaan, line 86: does it really store enums like this? Hmm

activity_calendar/models.py
```diff
@@ -22,45 +30,232 @@ class Meta:
...
+        if num_user_registrations is None:
+            num_user_registrations = self.get_num_user_subscriptions(user, recurrence_id=recurrence_id)
+        
+        if num_total_participants is None:
+            self.get_slots(recurrence_id=recurrence_id).aggregate(Count('participants'))['participants__count']
+        if num_max_participants is None:
+            self.get_max_num_participants(recurrence_id=recurrence_id)
+
+        # Can the user (in theory) join another slot?
+        user_can_join_another_slot = (self.max_slots_join_per_participant < 0 or \
```
helaan, line 176: If you are using -1 as a magic value, please do `== -1` and not `< 0` because this isn't really clear

activity_calendar/models.py
```diff
@@ -70,6 +265,14 @@ def clean_fields(self, exclude=None):
         if self.start_date >= self.end_date:
             errors.update({'start_date': 'Start date must be before the end date'})
 
+        if self.subscriptions_open < self.subscriptions_close:
+            errors.update({'subscriptions_open': 'Subscriptions must open before they can close'})
+
+        if self.subscriptions_open < timezone.timedelta():
```
helaan, line 271: what does `timezone.timedelta()` return?

activity_calendar/models.py
```diff
@@ -85,3 +288,93 @@ def clean_fields(self, exclude=None):
...
+    description = models.TextField(blank=True)
+    location = models.CharField(max_length=255, blank=True, null=True,
+        help_text="If left empty, matches location with activity")
+    start_date = models.DateTimeField(blank=True, null=True,
+        help_text="If left empty, matches start date with activity")
+    end_date = models.DateTimeField(blank=True, null=True,
+        help_text="If left empty, matches end date with activity")
+    
+    # User that created the slot (or one that's in the slot if the original owner is no longer in the slot)
+    owner = models.ForeignKey(User, on_delete=models.CASCADE, blank=True, null=True)
```
helaan, line 304: So if the owner of the slot deletes his account his past and future slots are deleted as well? Why not set blank?

activity_calendar/migrations/0002_auto_20200814_1431.py
EricTRL: Woops I forgot this was a new branch!

activity_calendar/models.py
EricTRL: Because that would mean we actually need to create them at some point in time. Requiring admins to create them means they need to create slots for each activity, which was a task we were wanting to avoid.
If the server needs to create them (in the DB), then it needs to be aware of when an activity's occurrence is and actually create the slots some time before that. Creating the slots when an activity is defined is not really an option, as activities can recur indefinitely.

Having the slots show up in the frontend without actually existing, and only creating them if they're being used, seemed like the easier approach.

---
This did make me realise that changing a recurring activity's start-date or recurrence-rules will break all already existing slots for that activity. The former is easy to fix; the latter not so much.

activity_calendar/models.py
EricTRL: `timezone.timedelta()` (without any parameters) returns a duration of 0 seconds. This statement basically checks for a negative duration.

activity_calendar/models.py
EricTRL: Good catch!

activity_calendar/models.py
EricTRL: This was changed in Django 3.0 but we're currently running 2.2 (the latest version of the 2nd major release)

activity_calendar/static/css/fullcalendar-customisation.css
EricTRL: I can't really remember, but they were roughly based on the bootstrap alert text colours (but do not seem to precisely match)

activity_calendar/models.py
EricTRL: I suggest we move a fix for that to a later branch that would already focus on supporting activity editing.

activity_calendar/forms.py
EricTRL: I honestly cannot remember why I decided to roll with this abomination.
Moved the relevant code to a logical place: the field's `clean` method

activity_calendar/models.py
helaan: Ok, I suggest we move that to the backlog and for now forbid editing recurrences

activity_calendar/models.py
helaan: Okay, then let's move upgrading to Django 3.0 to the backlog

activity_calendar/templates/activity_calendar/activity_slots.html
```diff
@@ -0,0 +1,210 @@
...
+            <strong class="alert-success">Hooray!</strong> You are currently registered for {{ num_registered_slots }} slot(s)!
+            <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+            </button>
+        </div>
+      {% endif %}
+
+      {% if deregister %}
+        <div id="info-msg" class="alert alert-success alert-dismissible fade show" role="alert">
+            <strong class="alert-success">Aww, why are you leaving?</strong> You successfully deregistered.
```
helaan, line 50: Please don't shame people for leaving

core/models.py
```diff
@@ -23,3 +26,23 @@ def get_simple_display_name(self):
     @staticmethod
     def set_display_name_method(method):
         ExtendedUser.display_name_method = method
+
+
+# File path to upload achievement images to
+def get_image_upload_path(instance, filename):
+    # Obtain extension
+    # NB: A file can be renamed to have ANY extension
```
helaan, line 34: Should there be a filter on the allowed file extensions? f.e. only .png, .jpg and .jpeg to prevent people from uploading .js files or other nasty stuff

core/templates/core/base.html
```diff
@@ -5,6 +5,7 @@
 <html lang="en">
 	<head>
 		<meta charset="UTF-8">
+		<meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1.0, maximum-scale=1.0">
```
helaan, line 8: Why did you disable zooming on all pages?

core/models.py
EricTRL: The relevant field is an ImageField which already does so :smile:
Do note that people can rename a file to have any extension though, and I'm not sure how this is handled.

helaan (approved):
Pull request is large and I'm not sure it is completely bug free, we'll just fix the bugs afterwards

helaan (approved):

PR #61, opened 2020-09-03
Fix/60 browser compatibility

### LINKED ISSUE #60: Browser Compatibility
Specifically the activity calendar suffers from issues regarding browser compatibility. Everything seems to work as expected in Google Chrome, but people have encountered issues in Edge and Firefox already.

Moreover, the mobile formatting could use a few improvements to increase both usability and compatibility. People have already reported issues with Firefox mobile.
From my own experience, the navbar doesn't really work on mobile browsers.

EricTRL:
Fixes include:
- MS Edge can display the activity calendar (formerly prevented by a JS error)
- Samsung Internet Browser can display the activity calendar (formerly prevented by the same JS error)
- The Navbar has a non-zero height on mobile devices (NB: Bug did not show on Google Chrome for desktop devices, even if the screensize was set to that of a mobile device using the developer panel)
- "Earned Achievements" page shows achievement icons
- One less missing closing `<div>`
- Left/right margin for content. This was only really noticeable on mobile devices.
- No fixes for Firefox Mobile as the current release seems to function correctly on it.

Closes #60

### DIFF
--- a/activity_calendar/static/js/fullcalendar-init.js
+++ b/activity_calendar/static/js/fullcalendar-init.js
@@ -10,7 +10,7 @@ document.addEventListener('DOMContentLoaded', function() {
             minute: '2-digit',
             meridiem: false, // AM/PM display
         },
-        initialView: 'timeGridWeek',
+        initialView: $(window).width() < 992 ? 'listWeek' : 'timeGridWeek',
         headerToolbar: {
             left: 'prev,next today',
             center: 'title',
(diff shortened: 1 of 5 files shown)


activity_calendar/static/js/fullcalendar-init.js
```diff
@@ -86,17 +86,11 @@ function onEventClick(info, calendar) {
...
-        if (!event.allDay) {
-            opts = {hour: 'numeric', minute: 'numeric'}
-        }
-
         date_str = start_date.toLocaleString('en-gb', {
-            month: 'long', day: 'numeric', ...opts
+            month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric',
         })
         date_str += ' - ' + end_date.toLocaleString('en-gb', {
-            month: 'long', day: 'numeric', ...opts
```
helaan, line 99: Maybe we should add an ES6 toolchain.

I can probably port that over from abot (I should maybe start porting over some stuff from abot anyway)

helaan (approved):

PR #63, opened 2020-09-07
Improve participants listing

helaan:
- Only show to logged in members
- Stop showing when the activity is over instead of registration close
- Deal with the colons

Also, extract to a seperate method

### DIFF
--- a/activity_calendar/models.py
+++ b/activity_calendar/models.py
@@ -162,6 +162,9 @@ class Activity(models.Model):
         # Slots cannot be created outside the admin panel; we have to work with
         # the slots that already exist
         return self.get_num_slots(recurrence_id=recurrence_id)
+
+    def get_duration(self):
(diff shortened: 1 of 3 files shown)


activity_calendar/models.py
```diff
@@ -240,6 +243,21 @@ def are_subscriptions_open(self, recurrence_id=None):
         now = timezone.now()
         return recurrence_id - self.subscriptions_open <= now and now <= recurrence_id - self.subscriptions_close
 
+    # Whether the name of participants is shown
+    def show_participants(self, user, recurrence_id=None):
```
EricTRL, line 247: NIT: Method name is somewhat misleading (as it only returns a boolean)

EricTRL (approved):

EricTRL (approved):

Issue #65, opened 2020-09-07
Membership file export


EricTRL:
Certain information of the membership file (such as email addresses or phone numbers) should be exported to a file when information changes. Depending on further server setup, it can then be shared using a service such as Nextcloud so that a (relatively) recent version is always available even if Squire's server is down.
Moreover, it should be possible to copy the list of member's email addresses.

PR #68, opened 2020-09-14
Feature/65 membership file export

EricTRL:
Upon saving, updating, or removing a member, certain information is stored in a CSV file for backup purposes.
The admin panel was made more usable (for the Member model and its related logs)

### DIFF
--- a/membership_file/admin.py
+++ b/membership_file/admin.py
@@ -1,11 +1,62 @@
 from django.contrib import admin
 from .models import Member, MemberLog, MemberLogField
-from django.contrib import admin
+from django.utils.html import format_html
+
+
+class HideRelatedNameAdmin(admin.ModelAdmin):
+    class Media:
+        # Hacky solution to hide the "X was updated: <Y> -> <Z>" text
+        # We don't need to edit this information anyways, so it's safe to hide
+        # https://stackoverflow.com/a/5556813
+        css = {
+            'all': ('css/hide_related_model_name.css',),
+        }
+
+class DisableModifications():
+    # Disable creation
+    def has_add_permission(self, request):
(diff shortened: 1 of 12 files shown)


membership_file/admin.py
```diff
@@ -1,11 +1,62 @@
 from django.contrib import admin
 from .models import Member, MemberLog, MemberLogField
-from django.contrib import admin
+from django.utils.html import format_html
+
+
+class HideRelatedNameAdmin(admin.ModelAdmin):
```
helaan, line 6: why this change instead of overriding the string representation?

membership_file/tests/tests_user.py
```diff
@@ -57,6 +57,7 @@ def test_field_label(self):
 ################################################################
 
 # Tests the editing feature of a user's own membership info.
+@override_settings(MEMBERSHIP_FILE_EXPORT_PATH=None)
```
helaan, line 60: why not test the export functionality instead?

helaan (approved):
Extra deployment steps required:
- merge settings.py (note: I should probably upstream these changes)
- run collectstatic as CSS was changed

membership_file/admin.py
EricTRL: It'd make debugging a lot trickier; printing/loggin the instance would then show nothing.
A non-hacky solution would be to modify tha admin template, but that seemed overkill.

membership_file/tests/tests_user.py
EricTRL: That's done in a different (and more general) testcase in `tests_export.py`

EricTRL:
Closes #65

PR #69, opened 2020-09-14
Feature/deregistered members

EricTRL:
Allows setting a member to 'deregistered'. This is necessary if (contact) information needs to be kept (E.g. if they still owe us money). A deregistered member sees themselves as a non-member in the front-end.

Rules for viewing who registered for an activity are more strict (only members can do so).

Also fixes a bug where a user could not be added to a member.


membership_file/admin.py
```diff
@@ -43,12 +43,12 @@ def get_url(self, obj):
 # Ensures that the last_updated_by field is also updated properly from the Django admin panel
 class MemberWithLog(HideRelatedNameAdmin):
 
-    list_display = ('id', 'first_name', 'tussenvoegsel', 'last_name', 'educational_institution', 'email', 'phone_number', 'marked_for_deletion')
-    list_filter = ['marked_for_deletion', 'educational_institution']
-    list_display_links = ('id', 'first_name')
+    list_display = ('id', 'user', 'first_name', 'tussenvoegsel', 'last_name', 'educational_institution', 'is_deregistered', 'marked_for_deletion')
```
ReneKOTKT, line 46: why are e-mail and phone number removed?

EricTRL: They took quite a bit of screen space, but did not really provide a useful sort or filter option.
They can still be viewed (or changed) on an individual member's page, so they're not removed entirely.

membership_file/tests/tests_admin.py
```diff
@@ -90,6 +90,7 @@ def setUpTestData(cls):
             "country": "",
             "member_since": "",
             "educational_institution": "",
+            "is_deregistered": "",
```
helaan, line 93: why is this not False?

helaan (approved):
Did not run it, seems sensible. Fixes a bug that you cannot link members to users

helaan (approved):
Eric, kun je in het vervolg bugfixes los pushen?

EricTRL:
> Eric, kun je in het vervolg bugfixes los pushen?

Ja dit was een foutje 😅

PR #75, opened 2020-09-23
fix/Need go register before you can join for activities where this is not required

EricTRL:
An activity for which registration was not required copied the popup "you need to register before you can join" from the last activity that was clicked on for which you do need to register.


helaan (dismissed):
Looked at it but did not test it, seems a little hacky. Preferred solution would be to always call show or hide instead of only calling show when needed.

Deploy note: Run collectstatic

helaan (approved):

Issue #66, opened 2020-09-10
settings do not define AUTH_USER_CLASS


DutcherNL:
Right now locally all references to users in code are replaced with the ExtendedUser proxy model. This makes maintaining code annoying and also requires a lot of arguably dubious user.__class__ = ExtendedUser when retrieving User from request (which is not the ExtendedUser.

It's better to make a non-proxy CustomUser(AbstractUser) class. However, this requires some rewriting in the database as this would break all currently automated intemediate model names.

DutcherNL:
See here: https://docs.djangoproject.com/en/3.1/ref/settings/#auth-user-model

It should not be adjusted after deployment, though the current implementation, I believe, will only become worse as we add more features, so we need to address this soon.

EricTRL:
`ExtendedUser` no longer exists as of #177, I don't think this is needed anymore?

PR #79, opened 2020-10-03
Updated activity display

DutcherNL:
A massive massive change in the activity views
No longer are errors hard coded in html and views, but instead Django Forms are used to check authorisation.
I made a small deviation from the Django Forms in order to adress certain denied options prior to user interaction (i.e. whether slots can be created before the user fills in the slot creation info).
All data validation and saving is done in the forms now instead of View
View is now responsible for displaying the data in a convenient way. As such the raised error in forms are converted to user directed texts instead of the non-directed error text created by the forms ("You can not... " instead of "User can not")

Still need to rewrite test_views to address the new view workings. The Form validation test cases have been written

Other improvements:
- Created a render_form template tag to render forms in bootstrap format (instead of the current HTML based render_field X method. (so we can add a field in the form and it's instantly reflected in all those form instances over the entire site).
- Addressed several direct url naming instead of relative view_name usage.
- Implemented Messaging framework
- Added contentFrame class to core/base instead of the many many many templates that inherent from it. If needed one can use {% block content-frame-class %} to change the content class name (or remove it)

Like I said in chat. I did a lot of different things


activity_calendar/templates/activity_calendar/activity_page_base.html
```diff
@@ -0,0 +1,203 @@
...
+                                width="16"
+                        />
+                    </div>
+                    <!-- Event Date & Recurrence Information -->
+                    <div class="col align-self-center">
+                        <div class='nospace'>
+                            <p id="event-date">
+                                {% if start_date.date == end_date.date %}
+                                    <!-- Activity starts and ends on the same day -->
+                                    {{ start_date|date:"l, j E H:i" }} - {{ end_date|date:"H:i" }}
```
helaan, line 62: 1. REMARK: I've checked the manual and the used parameters seem sensible: https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#date Maybe link these docs as the date formats aren't common across the code?
2. Why are you putting a comma between the weekday and day here, and between date and time on L65?

activity_calendar/templates/activity_calendar/activity_page_base.html
```diff
@@ -0,0 +1,203 @@
...
+                               class="btn btn-success btn-block">
+                                Log in
+                            </a>
+                        </div>
+                    </div>
+                </div>
+            {% endif %}
+            <div class="alert alert-warning">
+                Corona protocol is active! Got symptoms? Stay home!<br>
+                <a href="https://beta.kotkt.nl/nextcloud/s/JpGkgdMLCGX4nHL/download">
```
helaan, line 182: Outside this PR's scope: we should probably stop putting links to this protocol in the code

activity_calendar/templates/activity_calendar/activity_page_new_slot.html
```diff
@@ -0,0 +1,67 @@
+{% extends "activity_calendar/activity_page_base.html" %}
+{% load render_form %}
+{% load static %}
+{% load calendar_tags %}
+
+
+{% block XXX %}
```
helaan, line 7: is XXX a placeholder?

activity_calendar/templates/activity_calendar/activity_page_no_slots.html
```diff
@@ -0,0 +1,79 @@
...
+        <h2>Current participants</h2>
+        {% if participants %}
+            {% for participant in participants %}
+                <div class="">
+                    {{ participant.user.get_display_name }}
+                </div>
+
+            {% endfor %}
+        {% else %}
+            There are currently no people sign-ed up. {% if subscriptions_open %} Be the first! {% endif %}
```
helaan, line 76: I don't think there should be a dash between `sign` and `ed`

activity_calendar/templates/activity_calendar/activity_page_base.html
DutcherNL: O yeah, it was a temporary placeholder for early in the project. Gonna make a modal for that. Or at least the Corona rules (not the sign-up instructions, that should, if designed well, flow well)

activity_calendar/templates/activity_calendar/activity_page_new_slot.html
```diff
@@ -0,0 +1,67 @@
...
+
+{% block additional-contents %}
+    <hr>
+    <h2>Create new slot</h2>
+
+    <div class="wideContentFrame">
+        {% render_form id="create_slot_form" method="post" submit_class="btn btn-success btn-block" submit_text="Create slot" %}
+    </div>
+
+{% endblock %}
```
helaan, line 67: Please add a trailing newline.

_Looks at repo_

... why is there no editorconfig file in Squire yet. BRB.

membership_file/templates/membership_file/no_member.html
```diff
@@ -7,8 +7,9 @@
   Squire - Not a Member
 {% endblock title %}
 
+
+
```
helaan, line 11: NIT: why add empty lines?

core/templates/core/form_total.html
```diff
@@ -0,0 +1,27 @@
+{% load generic_field %}
+
+<form method="{{ form_method }}"
+      {% if form_id %}id="{{ form_id }}"{% endif %}
+      {% if form_id %}id="{{ form_id }}"{% endif %}
```
helaan, line 5: Why are you adding the id attribute twice?

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
```diff
@@ -0,0 +1,45 @@
...
+                {% endif %}
+                <p class="card-text">{{ slot.description }}</p>
+
+                <div class="row no-gutters">
+                    <div class="card-text">
+                        <small class="text-muted">Participants:
+                            {% if slot.max_participants != -1 %}
+                                {{ slot.participants.all|length }} / {{ slot.max_participants }}
+                            {% else %}
+                                {{ slot.participants.all|length }} / &infin;
```
helaan, line 30: Why show the infinity symbol here instead of omitting the limit completely if there is no limit?

So you'd have `Participants: 7` instead of `Participants: 7 / &infin;`

core/templates/core/form_total.html
DutcherNL: Don't know how that happened. Probably in a copy paste error somewhere, dunno. Fixed it

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
```diff
@@ -0,0 +1,45 @@
...
+                        <small class="text-muted">
+                            Alternative Location: {{ slot.location }}
+                        </small>
+                    </div>
+                {% endif %}
+                <p class="card-text">{{ slot.description }}</p>
+
+                <div class="row no-gutters">
+                    <div class="card-text">
+                        <small class="text-muted">Participants:
```
helaan, line 26: Why did you flip the order of Participant and the amount compared to the current version?

membership_file/templates/membership_file/no_member.html
DutcherNL: That probably happened accidentally when I was changing all templates that extend core/base.html to no longer contain the contentFrame attribute and I accidentally allocated the space for the frame processing.

helaan (commented):
I went through the HTML templates, may continue later with the python files

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
DutcherNL: Style continuity. Figured this might be a bit clearer that there is no cap without saying infinite particiapnts (which it said earlier)

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
DutcherNL: Because I had this in my code before your code was imported into my code :P Seriously, I was working and every now and then I'd see a message in the chat addressing some Issue that I already addressed without knowing it XD

helaan: I only added the list of participants in #63, the order of Participant and the amount did not change in that PR,

Reversing the order does solve the inflection issue that was present in the old version (1 Participants), but still I don't prefer this version, as I think that starting each line with "Participants: " together with the vertical looks a bit ugly as it doesn't form a correct sentence. What does Eric think?

EricTRL:
The popovers only close when clicking the same button again.
Could it be possible to make them dismissable: https://getbootstrap.com/docs/4.0/components/popovers/#dismiss-on-next-click

EricTRL (commented):
The popovers only close when clicking the same button again.
Could it be possible to make them dismissable? https://getbootstrap.com/docs/4.0/components/popovers/#dismiss-on-next-click

activity_calendar/templates/activity_calendar/activity_page_no_slots.html
```diff
@@ -0,0 +1,79 @@
...
+                        </button>
+                    {% endif %}
+                </div>
+            </div>
+        </div>
+    {% endif %}
+{% endblock sign-up-status %}
+
+{% block additional-contents %}
+    {% if request.user.is_authenticated %}
```
EricTRL, line 66: Currently, only members* can view who registered for a slot. This implicitly reverts that change.

*Members here means "users who are in Squire's membership file". Anyone can create a Squire account, and there's no need for (possibly) random people to see who's registered.

activity_calendar/templates/activity_calendar/googlecalendar.html
```diff
@@ -3,12 +3,15 @@
 {% load static %}
```
EricTRL, line 3: Out of scope for this PR, but this view can definitely be removed.

activity_calendar/templates/activity_calendar/slot_blocks/register_button.html
```diff
@@ -0,0 +1,25 @@
...
+{% if form.get_base_validity_error %}
+    {% if sign_up %}
+        <button type="button" class="btn btn-outline-success btn-block disabled btn-block" data-container="body" data-toggle="popover" data-placement="bottom"
+                data-content="{% form_rejection_reason form %}">
+            Register
+        </button>
+    {% else %}
+        <button type="button" class="btn btn-outline-danger btn-block disabled btn-block" data-container="body" data-toggle="popover" data-placement="bottom"
+                data-content="{% form_rejection_reason form %}">
+            Deregister
```
EricTRL, line 12: The 'Deregister' button displays when not logged in.

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
EricTRL: I agree with Henk. I'm personally not much of a fan of the vertical bar; it feels a little out of place.

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
```diff
@@ -0,0 +1,45 @@
...
+                            {% if slot.max_participants != -1 %}
+                                {{ slot.participants.all|length }} / {{ slot.max_participants }}
+                            {% else %}
+                                {{ slot.participants.all|length }} / &infin;
+                            {% endif %}
+                            {% if user.is_authenticated %} |
```
EricTRL, line 32: Same issue as mentioned earlier; it implicitly reverts a change where only members can view activity participants.
Furthermore, which users joined an activity can be seen even after said activity has ended. This is also not the case in the current codebase.

activity_calendar/templatetags/calendar_tags.py
```diff
@@ -0,0 +1,82 @@
...
+        initial = {
+            'slot_id': slot.id,
+            'sign_up': slot.get_subscribed_participants().filter(id=user.id).count() == 0
+        }
+    )
+    return form
+
+
+@register.simple_tag(takes_context=True)
+def get_opening_time(context, filter=None):
```
EricTRL, line 32: Nice addition! Might also be nice to have this on the activity calendar modal itself.

EricTRL, line 32: Also displays for past events whose subscriptions actually won't ever open again.

activity_calendar/templatetags/calendar_tags.py
```diff
@@ -0,0 +1,82 @@
...
+    open_date = context['start_date'] - context['activity'].subscriptions_open
+    if filter:
+        a = date_format(open_date, filter)
+        print(a)
```
EricTRL, line 36: 🤔

activity_calendar/views.py
```diff
@@ -131,171 +38,292 @@ def check_join_constraints(request, parent_activity, recurrence_id):
...
+            'participants': self.activity.get_subscribed_participants(recurrence_id=self.recurrence_id),
+        })
+        return kwargs
+
+
+class ActivityMomentWithSlotsView(LoginRequiredForPostMixin, FormMixin, ActivityMomentView):
+    form_class = RegisterForActivitySlotForm
+    template_name = "activity_calendar/activity_page_slots.html"
+
+    error_messages = {
```
EricTRL, line 182: `slot-full` should probably also be in this list; it currently displays "Undefined error occurred"

activity_calendar/views.py
```diff
@@ -131,171 +38,292 @@ def check_join_constraints(request, parent_activity, recurrence_id):
...
+        if self.activity.slot_creation == "CREATION_USER":
+            new_slot_form = RegisterNewSlotForm(
+                initial={
+                    'sign_up': True,
+                },
+                activity=self.activity,
+                user=self.request.user,
+                recurrence_id=self.recurrence_id,
+            )
+        elif self.activity.slot_creation == "CREATION_NONE" and self.request.user.is_staff:
```
EricTRL, line 218: Earlier on there was a superuser-check rather than staff check.
Moreover, staff members are not necessarily people who should also be able to interact with the activities. For instance, at the moment there is the 'Achievement Manager' role, which gives someone access to the achievements and claimants in the admin panel (and as such makes them a staff member).

(The same "TODO Permission system" comment as earlier also applies here)

activity_calendar/views.py
```diff
@@ -131,171 +38,292 @@ def check_join_constraints(request, parent_activity, recurrence_id):
...
+        # page anyway
+        error = context['form'].get_base_validity_error()
+        if error:
+            messages.error(self.request, self.error_messages.get(error.code, 'An unknown error occured'))
+            return HttpResponseRedirect(self.activity.get_absolute_url(self.recurrence_id))
+
+        return super(CreateSlotView, self).render_to_response(context, **response_kwargs)
+
+    def get_form_kwargs(self):
+        if self.activity.slot_creation == "CREATION_NONE" and self.request.user.is_staff:
```
EricTRL, line 287: Same comment as above.

core/templates/core/home.html
```diff
@@ -46,7 +44,7 @@ <h4>Future Plans</h4>
     <!-- Submit a Bug Report -->
     <h3>I found a bug!</h3>
     <p>Great! Or well...? Please notify the Board of what went wrong, or create an issue on <a href="https://github.com/EricTRL/Squire/issues">GitHub</a> if you know how that stuff works!</p>
```
EricTRL, line 46: Out of scope for this PR, but we should change the GitHub repository URL.

EricTRL:
All in all great changes! Definitely a major improvement over the current codebase 😄

activity_calendar/forms.py
```diff
@@ -1,41 +1,285 @@
...
+            slot_obj.participants.remove(self.user)
+            return False
+
+
+class RegisterNewSlotForm(RegisterAcitivityMixin, ModelForm):
 
     class Meta:
         model = ActivitySlot
-        fields = "__all__"
+        fields = ['title', 'description', 'location', 'max_participants']
```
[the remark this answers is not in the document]
DutcherNL: Yes, for now. We can still set them in the back-end, but most slots right now are made on the boardgame slot where only 1 was relevant. The only moment so far different images were used, from what I saw, was the OC activity. We can still adjust that in the back-end if wanted.
On longer term I might enable them for staff, but for now it was not that relevant imo

activity_calendar/templates/activity_calendar/activity_page_no_slots.html
DutcherNL: What's the way we can check on a user if that user is member?
And more importantly, does the request given member default to ExtendedUser or just User? Otherwise we might need to patch that in order to prevent code injection all over the place to interpret Users as ExtendedUsers (ugly), I made an issue about that a while ago I believe.

activity_calendar/templatetags/calendar_tags.py
DutcherNL: huh. So there was a print statement left. I thought I cleaned them all XD

activity_calendar/templates/activity_calendar/activity_page_no_slots.html
EricTRL: There's the `to_member` template filter which has been used in two other places, which converts a user (or extendeduser) to a MemberUser. A MemberUser is also an ExtendedUser so there's no need for _another_ (ugly) conversion.

There's indeed an open issue about it already (#66) and it is something we should fix.

activity_calendar/feeds.py
```diff
@@ -110,7 +109,9 @@ def item_timestamp(self, item):
 
     def item_link(self, item):
         # The local url to the activity
-        return item.get_absolute_url()
+        # Because of the repetitition of the activity there is not 1 date. So instead let's just reroute it to the
+        # app activity page for now
+        return "https://app.kotkt.nl/api/calendar/fullcalendar"
```
EricTRL, line 114: Hardcoded URL. You can try [this](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpRequest.build_absolute_uri) instead

activity_calendar/templates/activity_calendar/activity_page_base.html
```diff
@@ -141,7 +132,12 @@ <h1>{{ activity.title }}</h1>
                                     Activity is open for sign-ups
                                 {% endif %}
                             {% else %}
-                                Subscriptions will open on {% get_opening_time "j E, H:i" %}
+                                {% opens_in_future as opens_in_future %}
```
EricTRL, line 135: This seems like it shouldn't be there right?

activity_calendar/tests/test_views.py
```diff
@@ -0,0 +1,480 @@
...
+        self.assertIn('Terraforming Mars', slots)
+        self.assertIn('Ticket to Ride', slots)
+        self.assertIn('Pandemic', slots)
+        self.assertIn('Betrayal', slots)
+        self.assertIn('Boardgame the Boardgame', slots)
+
+        self.assertEqual(context['num_total_participants'], 3)
+
+    # Test POST without a correct url
+    # Even if the data is invalid, we expect a 400 bad request
```
EricTRL, line 144: Outdated comment

activity_calendar/feeds.py
DutcherNL: It'll be changed in a later PR that addresses feeds. This is a quick patch. Besides, from what I can tell Google doesn't use the link anyway

activity_calendar/templates/activity_calendar/activity_page_base.html
DutcherNL: It's a template tag, might have been better as a filter, I recognise that. I think I'll address that in the next PR that uses ActivityMoment objects to take care of these kind of things

core/templatetags/get_login_url.py
```diff
@@ -0,0 +1,29 @@
...
+from urllib.parse import urlparse, urlunparse
+
+from django.conf import settings
+from django.http import QueryDict
+from django.contrib.auth.views import redirect_to_login
+
+register = template.Library()
+
+##################################################################################
+# Template Tag that appends a * after a string if a given form field is required
```
EricTRL, line 11: I don't think this is the right description

core/templatetags/render_form.py
```diff
@@ -0,0 +1,31 @@
+from django import template
+from django.utils.translation import gettext_lazy as _
+
+register = template.Library()
+
+##################################################################################
+# Template Tag that creates standard form inputs based on a template
```
EricTRL, line 7: Same comment as above

DutcherNL: Ik moet meer opletten in mn copy pastes wanneer ik snel templates wil hebben in m'n code (ofwel de standaard imports en layouts enzo... XD

activity_calendar/templates/activity_calendar/googlecalendar.html
DutcherNL: It's been removed anyway

activity_calendar/templatetags/calendar_tags.py
DutcherNL: Can I forward that to a future PR?

EricTRL (approved):

PR #80, opened 2020-10-03
Add EditorConfig configuration file

helaan:
This makes it easier to share settings between editors and can serve as
a baseline while we don't have a formatter for that filetype installed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{css,html,js,json,py}]
+indent_style = space
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 
 [*]
 charset = utf-8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
```
helaan, line 7: This is the only line that needs to be discussed: currently all files in the repository use Unix-style (LF) line endings. Some Windows programs need Windows-style line endings (CRLF), so we may want to defer this to Git. Looking at the code Eric and Wouter checked in I see that they both use LF line endings in that code. 

In order to not break your workflow, please check if you guys have `git config core.crlf` enabled. If you have it enabled, this line should probably be removed

EricTRL: I think this is already handled by the .gitattributes
(`core.crlf` is disabled for me; `core.autocrlf` is enabled.)

https://github.com/esrg-knights/Squire/blob/0ef1125e313b98e2bc581c13141d631edab71939/.gitattributes#L1-L2

helaan: I didn't know you could do it that way, TIL, so I removed end_of_line from the config as this is then handled by Git.

EricTRL (approved):

PR #89, opened 2020-10-16
Custom activity moments

DutcherNL:
Added ActivityMoment model class that can locally overwrite certain information such as description or maximum number of guests
Redacted many methods in the Activity model class as they are either no longer used, or moved to activitymoment for more senseble use (e.g. because it required a recurrence_id)

ActivityMoment instances on the database are reflected in the calendar (onsite and the external one)
ActivityMoments can also be used to create unique one time events of an activity that are reflected in the calendar


activity_calendar/feeds.py
```diff
@@ -74,11 +89,19 @@ def vtimezone(self):
 
     def items(self):
         # Only consider published activities
-        return Activity.objects.filter(published_date__lte=timezone.now()).order_by('-published_date')
+        activities = Activity.objects.filter(published_date__lte=timezone.now()).order_by('-published_date')
+        exceptions = ActivityMoment.objects.all()
```
EricTRL, line 93: I don't think we should include activity moments for which their parent activity is not published yet.

activity_calendar/models.py
```diff
@@ -312,7 +206,137 @@ def get_absolute_url(self, recurrence_id=None):
...
+    last_updated = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        unique_together = ['parent_activity', 'recurrence_id']
+        # Define the fields that can be locally be overwritten
+        copy_fields = ['description', 'location', 'max_participants']
+        # Define fields that are instantly looked for in the parent_activity
+        # If at any point in the future these must become customisable, one only has to move the field name to the
+        # copy_fields attribute
+        link_fields = ['title', 'image']
```
EricTRL, line 278: I suggest we move title already to the customisable fields. It's not uncommon to have a shared boardgame night with another association and I don't think it's needed to require it to be a separate activity.

activity_calendar/models.py
```diff
@@ -312,7 +206,137 @@ def get_absolute_url(self, recurrence_id=None):
...
+    parent_activity = models.ForeignKey(Activity, on_delete=models.CASCADE)
+    recurrence_id = models.DateTimeField(verbose_name="parent activity date/time")
+
+    created_date = models.DateTimeField(auto_now_add=True)
+    last_updated = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        unique_together = ['parent_activity', 'recurrence_id']
+        # Define the fields that can be locally be overwritten
+        copy_fields = ['description', 'location', 'max_participants']
```
EricTRL, line 274: Setting an alternative start/end time might also be useful.

activity_calendar/migrations/0005_auto_20201016_1445.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.15 on 2020-10-16 12:45
```
EricTRL, line 1: Might as well squash these two migrations as they were introduced in the same branch and relate to the same model.

activity_calendar/tests/test_models.py
```diff
@@ -75,6 +169,10 @@ def test_clean_correct(self):
         self.base_activity.recurrences = deserialize_recurrence_test("RDATE:19700101T230000Z")
         self.base_activity.clean_fields()
 
+    def test_time_shift_in_get_occurences(self):
```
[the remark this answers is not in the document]
DutcherNL: I placed a todo there for a reason. But I didn't have the time to write it and a single test case is not that vital to checking overall code. Also get_occurences_between was already implemented earlier and thus not part of the PR. But is vital to have a testcase for (as it deals with the time issues taht are now tested in less logical places such as test_fullcalendar

activity_calendar/models.py
DutcherNL: I agree, but I'd rather see it in a future PR as different start/end times might bleed into a lot more places (such as sign-up deadlines)

EricTRL (approved):

PR #96, opened 2020-10-31
Front end activity moment editing

DutcherNL:
Added view for moment editing. Requires activity_model.can_edit permission for uses to access and edit the activity


EricTRL (commented):
image

Could be nice to have the activity's defaults as a HTML-placeholder (example in the activity's title above) rather than on the side.

activity_calendar/feeds.py
```diff
@@ -133,8 +133,9 @@ def item_end_datetime(self, item):
     def item_created(self, item):
         return item.created_date
 
+    @only_for(ActivityMoment)
```
EricTRL, line 136: Shouldn't activities without changed values also have a last-updated-date?

activity_calendar/templates/activity_calendar/activity_moment_form_page.html
```diff
@@ -0,0 +1,58 @@
+{% extends 'core/base.html' %}
+
+{% load static %}
+{% load render_form %}
+{% load generic_field %}
+{% load activity_default_display_tag %}
+
```
EricTRL, line 7: A title-block is missing

activity_calendar/templates/activity_calendar/activity_moment_form_page.html
```diff
@@ -0,0 +1,58 @@
...
+    wideContentFrame
+{% endblock %}
+
+{% block content %}
+    <h2>Change specific activity moment</h2>
+    <p>
+        Below you can edit attributes for this particular activity instance. Any field can be left empty. In that case
+        it will copy the attribute from the general activity instead.
+    </p>
+
```
EricTRL, line 18: It would be nice to have the date/time of the activity(moment) somewhere as well

activity_calendar/templates/activity_calendar/activity_moment_form_page.html
```diff
@@ -0,0 +1,58 @@
...
+{% load render_form %}
+{% load generic_field %}
+{% load activity_default_display_tag %}
+
+{% block content-frame-class %}
+    wideContentFrame
+{% endblock %}
+
+{% block content %}
+    <h2>Change specific activity moment</h2>
```
EricTRL, line 13: There's a mismatch between the terms "activity moment" and "activity instance". (I'd suggest we use "activity instance" or even "occurrence")

activity_calendar/templates/activity_calendar/activity_page_base.html
```diff
@@ -153,6 +153,14 @@ <h1>{{ activity_moment.title }}</h1>
         </div>
         <div class="col-12 col-md-5">
             {% if request.user.is_authenticated %}
+                {% if perms.activity_moment.can_edit %}
```
EricTRL, line 156: This permission does not exist in the database (it's not a Django default nor is this created in the ActivityMoment's Meta class)
I also suggest we use the term 'change' instead of 'edit' (this is done for other (Django) permissions as well)

activity_calendar/forms.py
```diff
@@ -288,3 +291,15 @@ def save(self, commit=True):
             slot_obj.participants.add(self.user)
 
         return slot_obj
+
+
+class ActivityMomentForm(ModelForm):
```
EricTRL, line 296: This creates an instance in the DB upon saving even if all (relevant) fields are left empty.
I'm also not sure if we want to delete an already existing activityMoments in case the user saves when all fields are empty.

activity_calendar/tests/test_models.py
```diff
@@ -174,6 +174,14 @@ def test_time_shift_in_get_occurences(self):
         pass
 
 
+class ActivityTestCase(TestCase):
+
+    def test_get_all_activity_moments(self):
```
EricTRL, line 179: This is left empty; is that intentional?

activity_calendar/views.py
```diff
@@ -365,3 +365,30 @@ def form_invalid(self, form):
...
+    def get_form_kwargs(self):
+        kwargs = super(EditActivityMomentView, self).get_form_kwargs()
+        kwargs.update({
+            'instance': self.activity_moment,
+        })
+        return kwargs
+
+    def form_valid(self, form):
+        form.save()
+        message = _("You have successfully changed the settings for '{activity_name}'")
```
EricTRL, line 389: I suggest we also include the activity's date/time here

activity_calendar/tests/test_form_validation.py
```diff
@@ -479,3 +481,59 @@ def test_save_form_creation_without_user(self, mock_tz):
...
+        form.save()
+        self.assertTrue(ActivityMoment.objects.filter(
+            parent_activity_id=1,
+            recurrence_id = "2020-08-28T19:00:00Z",
+        ).exists())
+
+    def test_validity_check(self):
+        """ Tests that normal form data is valid """
+        self.assertFormValid({
+            'local_description': 'Different title',
```
EricTRL, line 518: I don't think these should all be `local_description` 😉

activity_calendar/models.py
```diff
@@ -18,6 +18,8 @@
 # Models related to the Calendar-functionality of the application.
```
EricTRL, line 18: (Due to lack of a better place to put this)
The admin panel for ActivityMoment is not really user friendly, so it would be nice to include that in this PR as it's somewhat related and only a few lines.
There's also no __str__ method for ActivityMoment which is part of this problem.

EricTRL (changes requested):

activity_calendar/forms.py
DutcherNL: In the long term I think we want to link all slots an activity moment instance. So having an empty activityMoment instance is not a bad thing. In all relevant views we already initialise an activitymoment, it's just not present on the database.

activity_calendar/tests/test_models.py
DutcherNL: I'm currently writing that. Like I said in the chat. I'm still writing some additional test cases

activity_calendar/views.py
DutcherNL: Done

activity_calendar/admin.py
```diff
@@ -16,10 +16,28 @@ def is_recurring(self, obj):
...
+    def activity_moment_has_changes(obj):
+        """ Check if this ActivityModel has any data it overwrites """
+        for field in obj._meta.local_fields:
+            # Check for all local_ fields if it is None, if not, it overwrites an attribute
+            if field.name.startswith("local_"):
+                local_value = getattr(obj, field.name)
+                if local_value is not None and local_value != "":
+                    return True
+        return False
+    activity_moment_has_changes.short_description = 'Is tweaked'
```
EricTRL, line 35: We can use .boolean to have it display a checkmark/cross instead
```suggestion
    activity_moment_has_changes.boolean = True
    activity_moment_has_changes.short_description = 'Is tweaked'
```

activity_calendar/views.py
```diff
@@ -375,7 +375,13 @@ def form_invalid(self, form):
...
     form_class = ActivityMomentForm
     template_name = "activity_calendar/activity_moment_form_page.html"
-    permission_required = ('activity_moment.can_edit',)
+    permission_required = ('activity_calendar.change_activitymoment',)
+
+    def dispatch(self, request, *args, **kwargs):
+        for perm in request.user.user_permissions.all():
+            print(perm.codename)
+
+        return super(EditActivityMomentView, self).dispatch(request, *args, **kwargs)
```
EricTRL, line 384:
```suggestion
```

EricTRL (approved):

PR #98, opened 2020-11-04
Feature/private slot locations

### LINKED ISSUE #97: Hide alternative activity location from non-logged in users
Currently this is shown to everyone, which is an issue for activities hosted outside the university campus.

EricTRL:
Allows setting slot locations to "private". Private slot locations are only shown to users registered for that slot.

Closes #97


activity_calendar/models.py
```diff
@@ -356,13 +354,17 @@ def get_absolute_url(self):
...
         help_text="If left empty, matches start date with activity")
     end_date = models.DateTimeField(blank=True, null=True,
         help_text="If left empty, matches end date with activity")
 
+
+    location = models.CharField(max_length=255, blank=True, null=True,
+        help_text="If left empty, matches location with activity")
+    location_is_private = models.BooleanField(default=False,
+        help_text="Private locations are hidden for users not registered to this slot", verbose_name="private location?")
+
```
DutcherNL, line 367: While the request indeed does request that slots should be turned invisible, I have my doubts if we want it to be configurable on the slots themselves.
It would make more sense at the activitymoment or activity level itself so that all locations are possibly hidden until after sign-in or slot-registration (whatever is more favourable). It prevents mixed communications on a single page.

activity_calendar/templatetags/calendar_tags.py
```diff
@@ -50,12 +50,13 @@ def opens_in_future(context):
...
 def register_button(context, slot):
     user = context['request'].user
+    participants = context['participants']
 
     # There is no logged in user, so return nothing instead
     if not user.is_authenticated:
         return None
 
-    sign_up = not slot.participants.filter(id=user.id).exists()
+    sign_up = user not in participants
```
DutcherNL: To my knowledge the not in function uses python selection whereas the original query does searching on the database.
While current implementation does indeed save a database query (why this is an issue is beyond me) it does require participants to be in the local context when using this tag. That does not make sense (note that any call to context I made when writing these tags are guarenteed as they are covered in test cases). If you really want to save the database query this way, please instead require this tag to have participants as an input argument instead.

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
```diff
@@ -1,47 +1,56 @@
...
-            {% endif %}
+            <div class="col-12 col-md-2 d-flex align-items-center pt-3 pt-md-0">
+                {% if user.is_authenticated %}
+                    {% register_button slot %}
+                {% endif %}
+            </div>
         </div>
     </div>
-</div>
+{% endwith %}
```
DutcherNL: As stated in a private convo. The current patch has very little use. Sure it saves a query load, but only one. If you really want to optimise things here you may want to reduce the number of queries loading all particiapnt users when retrieving all slots with prefetch_related
https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-related
That could save a lot more than what you are trying to do here.

activity_calendar/models.py
EricTRL: Agreed! Will move to the activity(moment) model. 👍

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
EricTRL: prefetch_related (and its sibliing select_related) are definitely things we should look at someday, but I'm leaving that up for another (low-priority) PR

activity_calendar/models.py
```diff
@@ -259,6 +263,14 @@ def __new__(mcs, name, bases, attrs, **kwargs):
...
                         attrs[new_field_name] = new_field
+
+                        if isinstance(field, models.BooleanField):
+                            # Override choices for BooleanField with null=True.
+                            # Without this, the null-selection shows up as 'unknown' (which is misleading)
+                            new_field.choices = BLANK_CHOICE_DASH + [
+                                (True,  _('Yes')),
+                                (False, _('No')),
+                            ]
                         break
```
DutcherNL, line 274: I personally do not like this implementation, though I will except it, it would make much more create a widget for this and set this in the form instead (https://docs.djangoproject.com/en/3.1/ref/forms/widgets/). Note that widgets have a method `value_from_datadict(self, data, files, name)` that is called by the Form to translate the datadict (POST or otherwise) to the relevant.
Edit: There is a `NullBooleanSelect` widget that has an empty, yes and no option that it displays. This might be a lot better.

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
```diff
@@ -15,7 +15,15 @@ <h5 class="card-title">
                 {% if slot.location %}
                     <div class="row no-gutters">
                         <small class="text-muted">
-                            Alternative Location: {{ slot.location }}
+                            Alternative Location: {{ sign_up }}
```
DutcherNL, line 18: I can't find where {{sign_up}} is set, nor see why this should be here. Have you accidentally left this here?

activity_calendar/models.py
EricTRL: The NullBooleanSelect would actually be used without this snippet. However, this resulted in "Unknown" being the option for a null-value without any way to override this text.

https://github.com/django/django/blob/adb40d217ec57ade46b1394cfbf3c513dc669445/django/forms/widgets.py#L708-#L718

I do agree that a custom widget results in cleaner code.

DutcherNL: Ik zie twee opties:
1) Unknown word in de widget eerst door een gettext_lazy gegooid, dus je kan het in de vertaling overschrijven
2) Subclass NullBooleanSelect en in de init zet zelf self.choices in de init method (je kan een inheritence stap overslaan door `super(Select).__init__(attrs, choices)` of je kan na de super aan te roepen self.choices = [('unknown', 'Laat leeg'), ('true', 'Ja'), ('false', 'Nee') ]

activity_calendar/models.py
```diff
@@ -379,7 +383,7 @@ class ActivitySlot(models.Model):
     start_date = models.DateTimeField(blank=True, null=True,
         help_text="If left empty, matches start date with activity")
     end_date = models.DateTimeField(blank=True, null=True,
-        help_text="If left empty, matches end date with activity")
+        help_text="If left empty, matches end date with activity")
```
helaan, line 386: NIT: please configure your editor to remove trailing spaces

https://editorconfig.org/#download

activity_calendar/templates/activity_calendar/slot_blocks/slot_block.html
```diff
@@ -15,7 +15,15 @@ <h5 class="card-title">
                 {% if slot.location %}
                     <div class="row no-gutters">
                         <small class="text-muted">
-                            Alternative Location: {{ slot.location }}
+                            Alternative Location:
+                            {% if not activity_moment.private_slot_locations or user in slot.participants.all %}
+                                {{ slot.location }}
+                                {% if slot.location_is_private %}
```
helaan, line 21: I cannot find any other mention of `location_is_private` in this PR or in the codebase

EricTRL: And this is exactly why we have code reviews! It was not moved when the "private"-status of slots was moved to the activity itself.

HansNorg (approved):

PR #101, opened 2020-12-07
Set default country to NL

helaan:
As most members we have live there


membership_file/migrations/0003_auto_20201207_1432.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.17 on 2020-12-07 13:32
```
helaan, line 1: My venv shot itself because I upgraded my Python, so I have a more recent version than the rest, but that shouldn't cause issues. I have tested my changes.

DutcherNL (approved):

Issue #28, opened 2019-09-14
Membership File Validation


EricTRL:
Some fields are only required in some scenarios and some fields are mutually exclusive. E.g. Having a TU/e card number and external card number entered at the same time is unneeded, and a TU/e card number can only be entered if actually studying at the TU/e.

Can be fixed using model validation as mentioned in the [docs](https://docs.djangoproject.com/en/2.2/ref/models/instances/) or on [this](https://stackoverflow.com/questions/12639031/can-multiple-values-be-used-in-django-a-model-validator) StackOverflow thread.

DutcherNL:
What still needs to be validated what isn't yet validated?

EricTRL:
> What still needs to be validated what isn't yet validated?

I checked the model file, and the only that comes to mind is the fact that you'll need to leave a phone number when you have a keycode or external card (possibly tue student card).
From the user-side, this wouldn't change much (if they're keen on removing info they can add fake data anyways). For admins/the board, this can somewhat "ensure" that keycodes/cards are not given to people who are hard to reach.

PR #102, opened 2021-01-03
Feature/99 extra membership fields

EricTRL:
Added the requested fields (see the attached screenshot). Charfields that do not need to be unique can no longer be null (and instead default to the empty string). Most fields have become optional as they're not required in all cases. Validation for non-empty fields in specific cases (e.g. a phone number is required if a person as access to rooms) is not yet performed (see #28 ).

A new `Room` model was added to accommodate room access.

Refactored to class-based views to reduce the amount of code needed. Also moved last_updated_by logic to a single class.
Django's messaging framework is now also used when a user updates their own membership information.

<details>
  <summary>Screenshot (Admin Panel)</summary>

  admin-panel

  NB: A key-ID field is also included, but is not present in this screenshot.  
</details>

Note: This is still marked as a draft because #98 contains a change to how BooleanFields in forms are rendered.


DutcherNL (changes requested):
I appreciate the effort, lots of good and welcome changes. A couple of remarks that I'd like to see addressed. Bit of class renaming and clean-up of what I believe is unnecessary code cluttering but otherwise fine additions 👍 
I'm happy to see you use class based views. Cbv's are really a powerhouse in use and clarity imo once you learn how they work :)

DutcherNL (approved):

PR #108, opened 2021-03-17
Fix/rdate and exdate sometimes fail

### LINKED ISSUE #106: Specific-date recurrences can be off by one day in the calendar
When setting a specific recurring date in the admin panel, the activity shows up a day earlier in the calendar.

The cause is unknown at this point.

<details>
  <summary>Example Screenshot</summary>

  Admin Panel:
image

Calendar:
image
  
</details>

EricTRL:
- RDATEs are no longer off by one day when an activity occurs across a DST-switch.
- RDATEs show up at the appropriate time in external calendars, rather than at midnight (e.g. Google Calendar, Thunderbird, etc.)
- At most one RRULE can be included in each activity (RFC 5545)
- EXRULEs have been deprecated (RFC 5545)

Closes #106

### DIFF
--- a/activity_calendar/feeds.py
+++ b/activity_calendar/feeds.py
@@ -180,13 +180,13 @@ class CESTEventFeed(ICalFeed):
     @only_for(Activity)
     def item_rdate(self, item):
         if item.recurrences:
-            return list(util.set_time_for_dates(item.recurrences.rdates, item.start_date))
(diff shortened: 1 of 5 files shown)


helaan (approved):
Glanced over it, seems reasonable. I trust Eric enough to run this in production

PR #115, opened 2021-08-03
Fixed privacy issue for boardgame owners

DutcherNL:
Boardgame list is a public list. However it did display the full names of members who owned the game without needing to be a member. Patched the leak by checking if current (anonymous)user is a member


EricTRL (approved):
Ideally this'd work through a permission (like viewing achievement claimants), but this'll do for now.

PR #116, opened 2021-08-04
Inventory use improvement

DutcherNL:
Small bug fixes


inventory/admin.py
```diff
@@ -1,21 +1,35 @@
 from django.contrib import admin
 from django.contrib.contenttypes.admin import GenericTabularInline
+from django.contrib.contenttypes.forms import BaseGenericInlineFormSet
 
 from inventory.models import *
 
 
+class BaseGenericTweakedInlineFormSet(BaseGenericInlineFormSet):
+
+    def is_valid(self):
```
EricTRL, line 10: The `clean(..)` method might be a more suitable place for this setup?

EricTRL (changes requested):

inventory/admin.py
EricTRL: Actually, it's possible to get rid of this entire setup by overriding `Inventory.clean_fields(..)` instead and performing the content_object-existence check there. `content_type` and `object_id` are passed in that method's `exclude` parameter as they're not actually part of the form on the admin panel. See the at the bottom of [this](https://docs.djangoproject.com/en/3.2/ref/models/instances/#django.db.models.Model.clean) doc's paragraph.

```python
def clean_fields(self, exclude=None):
    super().clean_fields(exclude=exclude)
    if exclude and 'content_type' not in exclude and 'object_id' not in exclude:
        # Validate that content_object exists
        if self.content_object is None:
            raise ValidationError("The connected item does not exist", code='item_nonexistent')
```

core/templates/core/form_total.html
```diff
@@ -4,6 +4,7 @@
       {% if form_id %}id="{{ form_id }}"{% endif %}
       {% if form_action %}action="{{ form_action }}"{% endif%}
       {% if form_class %}class="{{ form_class }}"{% endif%}
+      enctype="{{ form_enctype }}"
```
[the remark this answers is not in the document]
DutcherNL: I've read up why the default is the default and I agree. Fixed it in latest commit.

inventory/admin.py
DutcherNL: That's a good point solution. I wasn't aware of that. That is indeed a lot better. Will implement it right away!

EricTRL (approved):

EricTRL (approved):

PR #118, opened 2021-08-07
Inventory maintenance improvement

DutcherNL:
Added improvements and additonal functionality to maintain item ownerships


inventory/templates/inventory/catalogue_for_type.html
```diff
@@ -90,4 +117,10 @@ <h1>Item catalogue for {{ item_type }}</h1>
 {% block js_bottom %}
     {{ block.super }}
     <script src="{% static "js/enlarge_item_image.js" %}"></script>
+
+    <script language="JavaScript">
```
EricTRL, line 121: Might as well make this a snippet in a separate template; it's being used in activity_page_base.html as well and I easily see it being used elsewhere in the future too.

inventory/tests/test_converter.py
```diff
@@ -0,0 +1,29 @@
+from django.test import TestCase
+from django.contrib.contenttypes.models import ContentType
+
+from inventory.urls import CatalogueConverter
+from inventory.models import *
+
+
+class TestCatalogueConverter(TestCase):
+    # fixtures = ['test_users', 'test_groups', 'test_members.json', 'inventory/test_ownership']
```
EricTRL, line 9: commented out code

inventory/templates/inventory/catalogue_for_type.html
```diff
@@ -37,33 +38,59 @@ <h1>Item catalogue for {{ item_type }}</h1>
...
                             Edit
                         </a>
-                    </td>
-                {% endif %}
-                <td class="w-50">
-                    {{ item.description }}
+                    {% endif %}
+                </td>
+                <td>
+                    {% if can_maintain_ownerships %}
```
EricTRL, line 60: Right now there are (at most) four buttons on a catalogue overview page, which is pretty crowded (edit item, edit ownership, add to committee, add to member). Furthermore, each of these buttons are also reachable from an item's link overview page.

I doubt it is needed to have shortcuts on such an overview page for stuff that won't be used daily. Hence, I suggest we replace all four of these buttons by a single "Edit" one that redirects to the link page, from where all the buttons can be accessed.

inventory/templates/inventory/catalogue_item_info_view.html
```diff
@@ -0,0 +1,171 @@
...
+                        >Edit</a>
+                        <form method="post" id="activate-{{ link.id }}"
+                              action="{% url "inventory:catalogue_item_link_activation" type_id=item_type item_id=item.id link_id=link.id %}">
+                            {% csrf_token %}
+                        </form>
+                        <input type="submit" form="activate-{{ link.id }}"
+                               class="btn btn-warning" value="Activate">
+                        <a href="{% url "inventory:catalogue_item_link_deletion" type_id=item_type item_id=item.id link_id=link.id %}"
+                           class="btn btn-danger">
+                            Delete
```
EricTRL, line 159: There's a bit of a mismatch(?) between where delete buttons are located: for boardgames they're in the edit page, but for ownerships they're only on the overview page.
I suggest we streamline this and move this "Delete' button also to the edit page. For active items, we can disable this button and add a tooltip stating that only inactive items can be deleted. I don't think such a note is anywhere at this point (or I completely missed it :P), but it might not be obvious to users that they cannot delete active items

EricTRL (changes requested):

inventory/templates/inventory/catalogue_for_type.html
DutcherNL: Die gebruikte popover, maar ik heb beide naar utils verplaatst nu als js-script.

inventory/templates/inventory/catalogue_for_type.html
DutcherNL: I agree that eventually we need to lessen it, but I want it availlable while we still migrate the massive catalogue of items. It saves a lot of clicks in the initial inventory of boardgame and roleplaying items

inventory/templates/inventory/catalogue_item_info_view.html
DutcherNL: Good point. I removed item removal from the edit page and to the maintenance page. Makes no sense to be able to remove it without knowing whom are the owners anyway

EricTRL (approved):

PR #138, opened 2021-09-03
Committees fix

DutcherNL:
Fixed issue occuring when checking membership of non logged-in users
Added option to add members to groups who are not a member (useful for old boards)


committees/models.py
```diff
@@ -71,7 +72,8 @@ class AssociationGroupMembership(models.Model):
     For another we want additional information stored that is difficult when attempting to override the link between
     user and group.
     """
-    member = models.ForeignKey(Member, on_delete=models.CASCADE)
+    member = models.ForeignKey(Member, on_delete=models.CASCADE, blank=True, null=True)
+    external_person = models.CharField(max_length=64, help_text="Person not is not a member", blank=True, null=True)
```
EricTRL, line 76: No need for `null=True` on a non-unique Charfield.
(there's also a double "not" in the help text)

EricTRL (changes requested):

EricTRL (approved):

PR #142, opened 2021-09-06
Member processing

DutcherNL:
Membership is now added to request through middleware and context through contextprocessor


EricTRL (approved):

membership_file/models.py
```diff
@@ -5,36 +5,24 @@
...
 # @since 06 JUL 2019
 ##################################################################################
 
 # Display method for a user that may also be a member
 def get_member_display_name(user):
-    member = MemberUser(user.id).get_member()
+    member = user_to_member(user)
     if member is not None:
         return member.get_full_name()
+    user.__class__ = User
```
EricTRL, line 20: Any particular reason why you're doing this? If yes, do add your reply as a comment in the code! :D

EricTRL (changes requested):

membership_file/models.py
DutcherNL: User is ExtendedUser instead of a normal User. I don't really know why we have a proxy user class for solely name displaying. This is something from the early days that I haven't been implementing in my new modules, but is currently still used in the membership and activity modules.

EricTRL (approved):

PR #153, opened 2021-09-11
Fix/activitymoment alt duration in calendar

EricTRL:
Activitymoments with a different start/end time should now always be included (or excluded) correctly from get_occurrences_between. This was not always done correctly before, but this likely went unnoticed.

Recurring activities occurring between 00:00-02:00 (in CEST) or 00:00-01:00 (CET) should now occur on the same day as they recurrence rule mentioned, rather than a day later. This was likely caused by a bug in django_recurrences, but it seems that we can work around it.


EricTRL:
Identified a few more issues while working on this:
- ActivityMoments in the snippet that allowed infinite participants, but which did not have required subscriptions, did not show their subscription status.
- Ensured that `Activity.recurrences` is not modified when calling `Activity.get_occurrences_between(..)`

(See the individual commit messages for a short list of modifications)

DutcherNL (approved):

PR #157, opened 2021-09-12
Added option in backend to add external users

DutcherNL:
> External user entries do not count as entry of that user
> Guests and users participants are counted seperately
> Added new total_participants parameter to activitymoment


activity_calendar/models.py
```diff
@@ -565,7 +580,12 @@ def get_absolute_url(self):
 class Participant(models.Model):
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     activity_slot = models.ForeignKey(ActivitySlot, on_delete=models.CASCADE)
+    # Charfield for adding external users, this can only be done through admin.
+    guest_name = models.CharField(max_length=123, default='')
```
EricTRL, line 584: Should probably add `blank=True` as empty values are allowed. The max length can also probably be lower

activity_calendar/models.py
```diff
@@ -421,8 +426,17 @@ def get_subscribed_users(self):
         return User.objects.filter(
             participant__activity_slot__parent_activity_id=self.parent_activity_id,
             participant__activity_slot__recurrence_id=self.recurrence_id,
+            participant__guest_name='',
         ).distinct()
 
+    def get_subscribed_guests(self):
```
EricTRL, line 432: There's a bit of a mismatch between the return values of `get_subscribed_users` and `get_subscribed_guests`. The latter returns `Participants`, the former `Users`. Possibly refactor to have `get_subscribed_users` also return Participants?

activity_calendar/models.py
```diff
@@ -403,6 +403,11 @@ def end_date(self):
         normal_duration = self.parent_activity.end_date - self.parent_activity.start_date
         return self.start_date + normal_duration
 
+    @property
+    def participant_count(self):
```
EricTRL, line 407: Property is not used in `get_json_from_activity_moment(..)` (activity_calendar -> api.py)

EricTRL (changes requested):

EricTRL (approved):

PR #176, opened 2021-11-19
Add prefers color scheme theme

LenaWil:
And fix a small bug with organisations being white on dark mode.

Okay, there is way to implement it with more JavaScript, which doesn't load the light theme, but that method got way too complicated way too fast, so I'm submitting this for code review instead.

### DIFF
--- a/core/static/css/main.css
+++ b/core/static/css/main.css
@@ -12,7 +12,7 @@ h1, h2, h3, h4, strong {
     color: var(--title-text-color);
 }
 
-.card, .modal-content {
+.card, .modal-content, .list-group-item {
     background-color: var(--body-alt-bg-color);
 }
 
--- a/core/static/css/main.css
+++ b/core/static/css/main.css
@@ -16,6 +16,10 @@ h1, h2, h3, h4, strong {
     background-color: var(--body-alt-bg-color);
 }
 
+.card-header {
+    background-color: var(--body-alt-dark-bg-color);
+}
+
 .icon {
     filter: var(--body-icon-color);
 }
--- a/user_interaction/static/themes/auto.css
+++ b/user_interaction/static/themes/auto.css
@@ -1,0 +1,2 @@
+@import url(./standard-theme.css);
+@import url(./dark-theme.css) (prefers-color-scheme: dark);
--- a/user_interaction/templates/user_interaction/home_users.html
+++ b/user_interaction/templates/user_interaction/home_users.html
@@ -113,7 +113,7 @@
                         we will send in our troops!
                     </p>
                     <p>
-                        <a class="btn btn-outline-dark" href="https://github.com/esrg-knights/Squire/issues"
+                        <a class="btn btn-outline-dark greenify" href="https://github.com/esrg-knights/Squire/issues"
                         target="_blank" rel="noreferrer noopener"
                         >Report Bug</a>
                     </p>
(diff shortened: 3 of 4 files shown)


LenaWil:
I didn't really know what to do with the current default theme and how to name everything, suggestions are welcome!

user_interaction/static/themes/auto.css
```diff
@@ -0,0 +1,2 @@
+@import url(./standard-theme.css);
```
LenaWil, line 1: I didn't succeed into adding a media query to this that also worked on Internet Explorer or other older browsers.

LenaWil:
And for those interested, [this branch](https://github.com/LenaWil/Squire/tree/themes/prefers-color-scheme/js-garbarge) contains some of the js code I was writing. 

It isn't really ordered though.

core/static/css/main.css
```diff
@@ -12,7 +12,7 @@ h1, h2, h3, h4, strong {
     color: var(--title-text-color);
 }
 
-.card, .modal-content {
+.card, .modal-content, .list-group-item {
     background-color: var(--body-alt-bg-color);
 }
```
EricTRL, line 18: Good fix! Perhaps we can also change the card header background colour so that it's slightly darker than the card body (just like in the light theme)
```suggestion
.card-header {
    background-color: var(--body-alt-dark-bg-color);
}
```

user_interaction/themes.py
```diff
@@ -21,15 +21,19 @@ def get_raw_js(self):
...
     css = ('themes/standard-theme.css',)
 
 class DarkUserTheme(SquireUserTheme):
     """ Dark Mode """
     name = "Dark Mode"
-    css = ('themes/dark-theme.css',)
+    css = LightUserTheme.css + ('themes/dark-theme.css',) #here, so it's consistent 
+
+class DefaultAutoTheme(SquireUserTheme):
+    name = "Automatically switch between dark and light"
```
EricTRL, line 35: Maybe just "Automatic" for the theme name?

user_interaction/static/themes/auto.css
EricTRL: I don't think we need to support old browsers like IE, as I doubt our members would still be using these

LenaWil: Okay, also all browser that don't support `prefers-color-scheme`, which still is 8% according to [caniuse.com](https://caniuse.com/?search=prefers-color-scheme).

user_interaction/templates/user_interaction/home_users.html
```diff
@@ -113,7 +113,7 @@ <h4>Encountered Evil?</h4>
                         we will send in our troops!
                     </p>
                     <p>
-                        <a class="btn btn-outline-dark" href="https://github.com/esrg-knights/Squire/issues"
+                        <a class="btn btn-outline-dark greenify" href="https://github.com/esrg-knights/Squire/issues"
```
LenaWil, line 116: It looked bad in dark mode, this was a quick fix.

EricTRL (approved):

LenaWil:
Screenshot 2021-11-20 at 16-23-10 Add prefers color scheme theme by LenaWil · Pull Request #176 · esrg-knights Squire

Am I supposed to not have write access?

PR #177, opened 2021-11-20
Refactor/ExtendedUser Removal + Admin Panel Changes

### LINKED ISSUE #139: Hidden fields in Admin inlines are displayed when using markdownform in admin view
The CSS used by markdown form overwrites the display: none feature of the django admin for the related field.

EricTRL:
This is a bit of a mishmash of changes to the admin panel. Furthermore, it removes the `ExtendedUser` Proxymodel that has been giving us a bit of a headache.

Complete Changelog:
- Added a new dev-dependency: [`django-debug-toolbar`](https://github.com/jazzband/django-debug-toolbar). This makes debugging easier, and can show things like template variables, django settings, or queries that are made when running the application with `DEBUG = True`
- Added `autocomplete_fields` to most ModelAdmins. These replace dropdown menus with special widgets that allow searching. As an added benefit, this (should) make these admin pages load faster.
- Removal of `ExtendedUser` and references to it. This was a proxymodel for Django's user object and basically only served as a wrapper to replace the standard user's `__str__` method. In this PR, we directly replace (or "patch") the user model's `__str__` method without introducing a new model or class. Although this does not replace the user class in Django settings (See #66), it does fix the workarounds we needed to use in order to actually use `ExtendedUser`. There's no need for database migrations as `ExtendedUser` was simply a proxy model for Django's standard user.
- Upgrades `django_dynamic_preferences` (fixes a bug)
- Closes #139 
- Added ModelAdmins for models that lacked one.
- Added Inlines for AchievementItemLinks
- Fixed a bug where MarkdownImages could not be uploaded for Systems or Committee-instructions.
- Changed `__str__` method of several models used with Inlines in the admin panel. They were not used in the frontend (nor will they ever be) and caused more DB-lookups that only really slowed down the page load.

Stale contenttypes (E.g. ExtendedUser which now no longer exists) can be removed from the production database using the `python manage.py remove_stale_contenttypes` management command.


achievements/tests/tests_serializers.py
```diff
@@ -27,7 +27,7 @@ def test_serializer_claimant(self):
             "extra_data_2":     "xyz",
             "extra_data_3":     "equal",
         })
-
```
LenaWil, line 30: Do we have a default formatter?

Then we can run that, add a test to ci with the formatter, and not have things like this anymore.

Then we don't get problems like this.

LenaWil:
Okay, I'm not familiar enough with code base to review this properly at the moment.

achievements/tests/tests_serializers.py
EricTRL: We've introduced a [`.editorconfig`](https://github.com/esrg-knights/Squire/blob/master/.editorconfig) at some point, but that wasn't there from the start. Changes like these basically mean that this file hasn't been touched since we introduced the editorconfig. Some IDEs support this out of the box, while others require an extension (like VSCode)

We haven't re-saved all the files in the repo to fix these issues in a single go, so that could be worthwhile to do.

membership_file/tests/tests_model.py
```diff
@@ -15,16 +15,16 @@ class MemberModelTest(TestCase):
...
+    # def test_member_display_method(self):
+    #     # Member has the correct display method
+    #     user = User.objects.filter(username="test_member").first()
+    #     display_str = get_member_display_name(user)
+    #     self.assertEqual(str(user), Member.objects.filter(user__username="test_member").first().get_full_name())
+
+    #     # Non-Member has the correct display method
+    #     user = User.objects.filter(username="test_user").first()
+    #     display_str = get_member_display_name(user)
+    #     self.assertEqual(display_str, user.get_simple_display_name())
```
DutcherNL, line 28: Is there a reason why it's commented out?

DutcherNL (approved):
Always a bit confusing if multiple functionality changes are implemented in the same PR. We (including myself) really need to stop doing that xD

membership_file/tests/tests_model.py
EricTRL: Past-me did not realise that present-me would forget to uncomment and update it 😅

DutcherNL: Figured, it's very familliar XD

DutcherNL (approved):

PR #186, opened 2021-11-24
Feature/assoc group activity pages

DutcherNL:
Add functionality to assoc_groups for activity pages and activity organisers

# Target goals
- [x] Add activity organisers functionality. They can do all admins can do on that activity
- [x] Add upcoming activity instances functionality on activities
- [x] Add assoc_group activity pages
- [x] Add create additional activity_moment functionality
- [x] Clearer repeating activity functionality
- [x] Update assoc_group config to include assoc_group permission access on tabs
- [x] Limit assoc_group activity page access through perms
- [x] Limit assoc_group activitymoment creation through perms

# Possible extensions
- [ ] Add create activity functionality
- [ ] Limit activity creation through perms
- [ ] Add an archived activity page to trace back older activities
- [ ] Add previous activitymoment page for activities to track past events


activity_calendar/tests/__init__.py
```diff
@@ -10,4 +10,12 @@ def mock_now(dt=None):
     def adjust_now_time():
         return timezone.make_aware(dt)
 
-    return adjust_now_time
\ No newline at end of file
+    return adjust_now_time
+
+
+def mock_is_organiser(result=True):
+    """ Script that changes the default now time to a preset value """
```
EricTRL, line 17: Comment and inner-method name seem outdated

activity_calendar/tests/test_forms.py
```diff
@@ -342,6 +342,16 @@ def test_slot_mode_admin_override(self, mock_tz):
         self.user.save()
         self.assertFormValid({'sign_up': True, 'title': 'My slot', 'max_participants': -1})
 
+    @patch('activity_calendar.models.Activity.is_organiser', side_effect=mock_is_organiser())
+    @patch('django.utils.timezone.now', side_effect=mock_now())
+    def test_slot_mode_organiser_override(self, mock_organiser, mock_tz):
+        """ Tests that form validates when an organiser creates a slot """
+        self.activity.slot_creation = Activity.SLOT_CREATION_STAFF
+        self.activity.save()
+        # self.user.user_permissions.add(Permission.objects.get(codename='can_ignore_none_slot_creation_type'))
```
EricTRL, line 351: 😄

activity_calendar/views.py
```diff
@@ -396,10 +405,12 @@ def form_invalid(self, form):
 # #######################################
 
 
-class EditActivityMomentView(LoginRequiredMixin, PermissionRequiredMixin, ActivityMixin, FormView):
+class EditActivityMomentView(LoginRequiredMixin, ActivityMixin, UserPassesTestMixin, FormView):
     form_class = ActivityMomentForm
     template_name = "activity_calendar/activity_moment_form_page.html"
-    permission_required = ('activity_calendar.change_activitymoment',)
+
+    def get_test_func(self):
```
EricTRL, line 412: I don't think this is needed anymore ;)

EricTRL (changes requested):
Awesome!

activity_calendar/views.py
DutcherNL: It is, it is used by UserPassesTestMixin to stop access to this particular page if can_edit_activity returns false. This is not default behaviour from the ActivityMixin

activity_calendar/views.py
```diff
@@ -90,6 +90,10 @@ def get_context_data(self, **kwargs):
 
         return kwargs
 
+    def can_edit_activity(self):
```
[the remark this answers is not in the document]
DutcherNL: I'll keep it here as it is behaviour specificially for the view. The model should not be aware which permissions it needs to access certain views/actions and it is not used elsewhere than the views.

activity_calendar/committee_pages/views.py
```diff
@@ -0,0 +1,64 @@
...
+        context = super(ActivityCalendarView, self).get_context_data(
+            tab_selected='tab_activity',
+            can_add_activitymoments=self.association_group.site_group.permissions.
+                                        filter(codename="add_activitymoment").exists(),
+            **kwargs,
+        )
+
+        return context
+
+class AddActivityMomentCalendarView(AssociationGroupMixin, FormView):
```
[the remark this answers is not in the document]
EricTRL: I'd rather stick to the RFC as we're also exporting our calendar, which is imported into other programs that (should) adhere to that same RFC.

I agree though that we do want to be able to create activitymoments outside of the recurrence pattern. We can still adhere to the RFC if we add an RDATE to the activity's recurrence pattern for the new activitymoment we create. Doing so should also make the activity visible in the calendar again.

I'm fine with waiting for a future PR.

EricTRL (approved):
As discussed in a PM, there was a different vision on Activities and how ActivityMoments outside the recurrence scheme should be handled. We settled on allowing ActivityMoments to appear outside their parent's recurrence scheme within Squire.
As a result, their ical export should be adapted to transform them into a "normal" activity (by copying their parent's info & removing their recurrence id). This way, the RFC specifications are still followed.
Likewise, we should adapt `Activity.get_activitymoments_between` to include ActivityMoments outside of their parent's recurrence scheme.

PR #188, opened 2021-11-26
Feature/homepage pins

EricTRL:
Allows pins to be added to the homepage. Pins can be linked to a specific model instance that inherits from `PinnableMixin`, allowing the pin to copy over certain attributes of the inheriting model.

Summary:
- Pins appear on both the authenticated and unauthenticated homescreens.
- Pins can be set to "members-only", meaning they will only display to those with a specific permission (which can be granted to, say, all members).
- Pins feature a title, description, url, and image. They also have a publication date and can have an expiration date.
- Models that implement the `PinnableMixin`-class can be linked to a pin. Attributes can then be automatically copied over from that model to a pin. Changes in the model will also be reflected in the pin. Pinnable models include: ActivityMoments, ActivitySlots, Committees, Items (Boardgames, Roleplaying Items, Miscellaneous Items), and Systems.
- Added a "Pin" button on each ActivityMoment page, which will pin/unpin that activitymoment. If no ActivityMoment exists for that page yet, it will be automatically created (as it's needed for the pin). ActivitySlots can also be pinned through the front-end on the same page. All other Pinnable models don't have such a button, but can still be added through the admin panel if desired.
- Pins use a specific template, but Pinnable models can override this template. This is currently not done, however.
- If a pin's description exceeds a certain length, users can "Read More" in a modal. This modal is added in the pin template.

TODOs:
- [ ] Testcases
- [ ] Pin History page. So far at most 3 pins can be displayed at the same time.


DutcherNL (commented):
Hey Eric, I looked over the code and while I really appreciate this feature, I disagree with the way you handle PinnableModelMixin though. The functionality introduced in this Mixin is bloating a lot of the model class with functionality that are not relevant for the model instance. The only functionality that makes sense is the GenericRelation and the option to detect (though from what I can tell this is not yet implemented) whether the linked content_type is valid for the pin.

I'd much rather see an additional class be created called `PinVisualiserBase` with all relevant visual information such as the templatename, pin_message_name, pin_message_title and all other retrievable parameters. You can subclass this for the relevant objects (`ActivityMomentPinVisualiser`, `BoardgamePinVisualiser` etc) with the desired pins.
In PinneableModelMixin you can define pin_visualisation_class for those items and when any pin needs to be rendered call the pinneablemixin-objects `get_visualiser()` in order to get the related visualiser object.

I think that is a much cleaner way to maintain pinneable objects that is also easier to maintain (as in, all display logic for a pin is in a class unrelated to database connection)

EricTRL:
I'm closing this as I'm unhappy with the current setup. There's a lot of code for pins in many different places, often closely knit to the database setup if an object instance needs to be linked. It also does not help that there's no clear view in what pins are supposed to be. Just a link to another page/annotated shortcut? An announcement?

I feel like it would be better to rebuild this from the ground up with a fresh new look, but only after it's clear what a pin is supposed to do.

PR #189, opened 2021-11-27
Fix calendar in Nemesis and Kinjin theme

LenaWil:
The Q theme was replacing the entire body, so we assumed it was okay for Kinjin.

It wasn't.

Also fixes the calendar in the Q theme.

### DIFF
--- a/user_interaction/static/themes/q-theme.js
+++ b/user_interaction/static/themes/q-theme.js
@@ -1,9 +1,9 @@
+"use strict";
+
 // Modify the DOM
-$( document ).ready(function() {
-    console.debug('hello!')
-    let html = $('body main').html()
+$(document).ready(function () {
+    console.debug('hello!');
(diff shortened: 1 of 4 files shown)


user_interaction/static/themes/q-theme.js
```diff
@@ -1,9 +1,9 @@
+"use strict";
+
 // Modify the DOM
-$( document ).ready(function() {
-    console.debug('hello!')
-    let html = $('body main').html()
+$(document).ready(function () {
+    console.debug('hello!');
     // Replace 'k' and 'c' by 'q' (leave html between < and > alone)
```
LenaWil, line 6: Strictly speaking the complicated regex _should_ not be needed anymore. But it's scary to touch, and doesn't hurt anyone.

Ndoll1 (approved):
👍🏻

PR #197, opened 2021-12-01
Codecleaning/account pages

DutcherNL:
Add account pages through automatic config settings in modules. Similarly to what is allready happening with committees

Todo:

- [x] Write config detection
- [x] Implement Membership
- [x] Implement Group overview
- [x] Implement achievements
- [x] Implement inventory
- [x] Implement site behavior (User account and preferences)

Bottom three are addressed in #201 for readability of the changes
- [x] Create ordering parameter
- [x] Create config validity check (is or is not member)
- [x] Optimise config storage through cache


achievements/account_pages/config.py
```diff
@@ -0,0 +1,16 @@
...
+class AchievementConfig(AccountConfig):
+    url_keyword = 'achievements'
+    tab_select_keyword = 'tab_achievements'
+    name = 'My achievements'
+    url_name = 'achievements'
+
+    def get_urls(self):
+        """ Builds a list of urls """
+        return [
+            path('', AchievementAccountView.as_view(), name='achievements'),
```
EricTRL, line 15: There's a bit of an overlap between `tab_select_keyword` and `AchievementAccountView.selected_tab_name`. Perhaps pass it in as a kwarg in as_view so it's in a single place?
```suggestion
            path('', AchievementAccountView.as_view(selected_tab_name=self.tab_select_keyword), name='achievements'),
```
The same goes for the accountconfigs of the other modules.

achievements/account_pages/views.py
```diff
@@ -0,0 +1,27 @@
...
+from django.views.generic import TemplateView
+
+from membership_file.util import MembershipRequiredMixin
+from user_interaction.account_pages.mixins import AccountTabsMixin
+
+from achievements.models import Category
+from achievements.serializers import CategorySerializer, AchievementSortType
+
+
+class AchievementAccountView(MembershipRequiredMixin, AccountTabsMixin, TemplateView):
```
EricTRL, line 11: Seems like the test for this view was removed (see `achievements/tests/tests.py`). I don't really mind though given that this module should be refactored anyways.

EricTRL: I don't see why membership is required for this view? My intentions with this Mixin were to limit view accessibility to members where it would not actually be possible for non-members to go regardless of permissions (e.g. editing membership information, which would not be possible for non-members).
In my opinion, things that are theoretically possible by non-members should be restricted using permissions if they should not be allowed to go there. I don't see why we would need to restrict a access to user's own earned achievements though (this is currently not the case either).

committees/account_pages/config.py
```diff
@@ -0,0 +1,19 @@
+from django.urls import path
+
+from user_interaction.config import AccountConfig
+
+from .views import AssociationGroupAccountView
+
+
+
+class TestAccountConfig(AccountConfig):
```
EricTRL, line 9: Might want to update this classname ;)

committees/account_pages/views.py
```diff
@@ -0,0 +1,12 @@
+from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.views.generic import TemplateView
+
+from membership_file.util import MembershipRequiredMixin
+
+from user_interaction.account_pages.mixins import AccountTabsMixin
+
+
+class AssociationGroupAccountView(MembershipRequiredMixin, AccountTabsMixin, PermissionRequiredMixin, TemplateView):
```
EricTRL, line 9:
```suggestion
class AssociationGroupAccountView(AccountTabsMixin, PermissionRequiredMixin, TemplateView):
```
The template used here contains a specific message if the user is not a member, so non-members can access this page too at the moment.

core/templates/core/base.html
```diff
@@ -95,12 +95,12 @@
 								Account
 							</a>
 							<div class="dropdown-menu dropdown-menu-right greenify-dropdown" aria-labelledby="navbarDropdownMenuLink">
```
EricTRL, line 97: We could possibly populate this dropdown using the config as well in a future PR.

user_interaction/account_pages/config.py
```diff
@@ -0,0 +1,20 @@
+from django.urls import path
+
+from user_interaction.config import AccountConfig
+
+from .views import SiteAccountView, AccountPasswordChangeView, LayoutPreferencesUpdateView
+
+
+class TestAccountConfig(AccountConfig):
```
EricTRL, line 8: Might want to rename this class ;)

user_interaction/config.py
```diff
@@ -0,0 +1,46 @@
+from importlib import import_module
+
+from django.apps import apps
+
+
+class AccountConfig:
+    """ Configurations for additional tabs on membership pages """
```
EricTRL, line 7:
```suggestion
    """ Configurations for additional tabs on user account pages """
```

user_interaction/config.py
```diff
@@ -0,0 +1,46 @@
...
+        > path('url_path/', views.MyMemberTabView.as_view(config=MyMemmberConfig))
+        """
+        raise NotImplementedError
+
+    @property
+    def urls(self):
+        return self.get_urls(), self.namespace, self.namespace
+
+
+def get_all_configs():
```
EricTRL, line 29: Bit of an overlap with the committee-configs, but that's fine for now.

user_interaction/templates/user_interaction/account_pages/site_account_page.html
```diff
@@ -18,6 +28,16 @@ <h1>Account</h1>
         <small class="text-muted">This information can currently not be edited on your end. Please notify the <a href="mailto:uups@kotkt.nl">UUPS</a> if something needs to be changed.</small>
     </p>
     <p>
-        You can change your password <a href="{% url 'core:user_accounts/password_change' %}">here</a>
+        You can change your password <a href="{% url 'account:password_change' %}">here</a>
     </p>
+
+    <h3>Look and layout</h3>
+    <p>
+        Site theme: {{ user.preferences.layout__theme }}
```
EricTRL, line 36: This shows the internal name of the theme instead of the human-readable variant. Fine enough for now.

utils/testing/view_test_utils.py
```diff
@@ -53,9 +54,14 @@ def assertValidPostResponse(self, data=None, url=None, redirect_url=None, fetch_
         data = data or {}
         response = self.client.post(url, data=data)
         if redirect_url:
+            # If a form errors, it returns a templateresponse instead. So we can instantly debug
```
EricTRL, line 57: Should probably add `utils/testing` to the `.coveragerc`: Coverage.py is picking up missed lines for testing utilities.

EricTRL (changes requested):

achievements/account_pages/views.py
DutcherNL: 1) Coherence, we do not support this page code-wise for non-members so we shouldn't allow this page linked like this
2) We should ideally have achievements connected to members, not users. In fact, I thought it was connected to members, not users.

EricTRL (approved):

PR #211, opened 2021-12-06
Oops forgot something, dark theme fix

Ndoll1:
"You are not registered for this activity
Claim a slot below"

Now on red background

### DIFF
--- a/user_interaction/static/themes/dark-theme.css
+++ b/user_interaction/static/themes/dark-theme.css
@@ -65,3 +65,7 @@ a:hover {
     color: white;
     background-color: red;
 }
+
+.bg-light {
+    background-color: red !important;
+}
--- a/user_interaction/static/themes/dark-theme.css
+++ b/user_interaction/static/themes/dark-theme.css
@@ -66,6 +66,6 @@ a:hover {
     background-color: red;
 }
 
-.bg-light {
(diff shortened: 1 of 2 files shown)


user_interaction/static/themes/dark-theme.css
```diff
@@ -65,3 +65,7 @@ a:hover {
     color: white;
     background-color: red;
 }
+
+.bg-light {
```
DutcherNL, line 69: This results in any bg-light to turn red and we can not guarantee that that is the case. It makes more sense to treat bg-light in a dark color scheme as a darker instance than bg-dark making it pop more with the light text.
Or, if you specifically want to target the alert require a combination of both bg-light and border-danger for this to trigger:
```suggestion
.bg-light.border-danger {
```

user_interaction/static/themes/scala-theme.css
```diff
@@ -66,3 +66,7 @@ a:hover {
     color: white;
     background-color: red;
 }
+
+.bg-light {
```
DutcherNL, line 70: See other comment
```suggestion
.bg-light.border-danger {
```

DutcherNL (changes requested):

DutcherNL (approved):

PR #212, opened 2021-12-06
Feature/activity cancellation

DutcherNL:
Add activity cancellation as an option.
Currently only possible on activitymoments from the admin:

status = Cancellation
cancels an activity, but keeps it visible on the site with the relevant information.
Marks calendar feed on activity as CANCELLED (should have effect on calendar software, though don't know exactly how yet)

status = Removed
Removes the activity from all availlable information streams, but keeps it present on the database.
Also sets the activitymoment on the calendar feed in the EXDATE so it doesn't show up at all.


Ndoll1 (approved):
Found no issues running the code, could cancel the activity. Code looks fine

PR #218, opened 2021-12-14
Update/inventory value system

DutcherNL:
Added option value functionality for association materials


Ndoll1 (changes requested):
Can't change value in the front end after linking the item to the committee, plz do fix :P

Also value of boardgames fluctuates, do you mean initial purchase price or really the current value. If it the latter we can change it almost weekly. Current name is ambiguous without extra explanation

Ndoll1 (dismissed):

inventory/admin.py
```diff
@@ -27,6 +30,41 @@ class OwnershipAdmin(admin.ModelAdmin):
...
+
+    def dehydrate_item_type(self, ownership):
+        return str(ownership.content_object.__class__.__name__)
+
+    def dehydrate_item_name(self, ownership):
+        return str(ownership.content_object.name)
+
+
+@admin.register(OwnershipValueProxy)
+class OwnershipValues(ExportActionMixin, admin.ModelAdmin):
```
EricTRL, line 55: As far as I can see, `OwnershipValueProxy` is only used to be able to create a second "ownership" management page in the admin panel, but with a limited queryset (being items owned by the association). Furthermore, I doubt we will ever want to export a subset of owned items, so the selection of items through an admin action seems unnecessary.

Instead, we could export via a [list filter](https://django-import-export.readthedocs.io/en/latest/getting_started.html#exporting-via-list-filters) and limit the queryset through `ExportMixin.get_export_queryset(..)` or even through the resource_class's `get_queryset(..)`. This way, the proxy model can be removed and all relevant information can be kept on a single admin page.

inventory/migrations/0001_initial.py
```diff
@@ -32,10 +33,14 @@ class Migration(migrations.Migration):
             name='Ownership',
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('added_since', models.DateField(default=datetime.date.today)),
+                ('added_since', models.DateField(default=django.utils.timezone.now)),
```
EricTRL, line 36: I don't think we should modify migrations that are already live in production?

EricTRL (changes requested):

inventory/migrations/0001_initial.py
DutcherNL: Agreed, except that this file shouldn't exist at all as it was used in the squashedmigration. But it was incorrectly left here and as a result I couldn't make a single change to Ownership as the intermediate files have been removed (this problem caught my attention last monday when I spend nearly two hours figuring out why the model registry failed when trying to migrate). If you want to you can try to recreate it on a clean branch and see if you don't have any issues with it on your local database, but this was the solution I came up with to fix the migration registry

inventory/admin.py
DutcherNL: What I tried to do, though I admit there are probably better ways than this proxy model hack, is make an additional Admin section that is focussed solely on the value of the inventory and not all other attributes (and additionally limited to only association owned items as only those values count). I think this might be a bit better for active content management, enabling the board to only access this list and not the complete list of options including contenttype and contentid.

inventory/admin.py
```diff
@@ -27,6 +30,41 @@ class OwnershipAdmin(admin.ModelAdmin):
 admin.site.register(Ownership, OwnershipAdmin)
 
 
+class OwnershipValueProxy(Ownership):
```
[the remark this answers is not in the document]
DutcherNL: Migrations aren't necessary as this function only as a stand-in shell and should and does not do anything on the database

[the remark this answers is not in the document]
EricTRL: The main problem I have is that migrations will be created for it anyway if, at some point in time, we run `python manage.py makemigrations`, even if that's done in a completely unrelated feature branch. It's possible to only run `makemigrations` for a specific app to avoid this, but that sounds like a workaround that someone is going to forget to do at some point.

inventory/admin.py
EricTRL: Even though there's still a lot of duplicate info, I suppose that's fair since the board will only be able to access the proxy model. I'm not entirely convinced with the current setup, but on the other hand the admin panel is quite finnicky to work with when wanting to do something non-standard. We might revisit it _at some point_ in the future if we encounter a better setup, but it'll definitely do for the time being.

EricTRL (approved):

PR #222, opened 2021-12-20
Feature/memberyears

DutcherNL:
Introduced MemberYears which enables linking members to  specific years, allowing for more structured transition through the years.
Also added front-end functionality for members to inform Squire they intend to stay member for the next year. This is done through the year-for-signup-promotion in global preferences. It automatically displays extend membership notification on front  page and membership profile page.

To do:

- [x] Handling for accounts that are not a member or no longer a member
- [x] Write testcases
- [x] Confirm written code for admin action to set user to active year
- [x] Update RequiresMembershipMixin for views on the correct places (i.e. limit access to users with connected member details, but who are not a member in the active  year(s)

Note: It is possible to have more than one active year running to allow overlap between the two years (e.g. in september)


membership_file/dynamic_preferences_registry.py
```diff
@@ -0,0 +1,21 @@
...
+from dynamic_preferences.preferences import Section
+from dynamic_preferences.registries import global_preferences_registry
+
+from membership_file.models import MemberYear
+
+membership_section = Section('membership')
+
+
+@global_preferences_registry.register
+class PromteSignUpYear(ModelChoicePreference):
```
EricTRL, line 11: Typo in the classname?

membership_file/templates/membership_file/extend_membership.html
```diff
@@ -0,0 +1,25 @@
...
+{% endblock title %}
+
+{% block og-title %}
+    Squire - Extend membership
+{% endblock og-title %}
+
+{% block content %}
+    <h1>Continue your quest</h1>
+    <p>
+        Another new year awaits filled with new adventures in distance realms and beautiful victories. Do you want to
```
EricTRL, line 16:
```suggestion
        Another new year awaits filled with new adventures in distant realms and beautiful victories. Do you want to
```

membership_file/templates/membership_file/extend_membership_successpage.html
```diff
@@ -0,0 +1,34 @@
...
+        that you will continue your quest through these treachures lands!
+    </p>
+
+    <p>
+        Please do note that a payment is requested post haste! You can transfer your contribution fee of
+        <b>&euro;25</b> to the bankaccount of the Knights:
+    </p>
+    <p><b>
+        NL49 RABO 0325427771
+    </b></p>
```
EricTRL, line 26:
```suggestion
    <p>
        The pigeon has been send out to report to the rulers of your kingdom. They will surely be most pleased
        that you will continue your quest through these treacherous lands!
    </p>

    <p>
        Please do note that a payment is requested post haste! You can transfer your membership fee of
        <b>&euro;25</b> to the bank-account of the Knights:
    </p>
    <p><b>
        NL49 RABO 0325 427 771
    </b></p>
    <p><small>Alternatively, you can pay in cash at the Knights' treasurer.</small></p>
```

membership_file/tests/tests_model.py
```diff
@@ -24,10 +25,44 @@ def test_memberlogfield_display(self):
         memberlogfield = MemberLogField(id=2, field="name", old_value="Bob", new_value="Charlie")
         self.assertEqual(str(memberlogfield), f"name was updated: <Bob> -> <Charlie> (2)")
 
+    def test_is_considered_member_active_years(self):
+        """ Tests member.is_considered_member when active years are present """
+        self.assertTrue(Member.objects.get(id=1).is_considered_member())
+        self.assertFalse(Member.objects.get(id=2).is_considered_member())
+        # Member is deregistered so should not be member, even though there is a active membesrhip connected
+        # (why this scenario would occur in real life I don't know, but let's set consistent behavior)
```
EricTRL, line 33: I suppose it could be possible that someone explicitly deregisters halfway through the year.

membership_file/util.py
```diff
@@ -50,12 +50,13 @@ class MembershipRequiredMixin(LoginRequiredMixin):
...
     """
     fail_url = settings.MEMBERSHIP_FAIL_URL
+    requires_active_membership = True  # Boolean defining whether user should be active, or just linked as an (old) member
 
     def dispatch(self, request, *args, **kwargs):
         if request.member is None:
             # Current session has no member connected
             return HttpResponseRedirect(resolve_url(self.fail_url))
-        if not request.member.is_considered_member():
+        if not request.member.is_considered_member() and self.requires_active_membership:
```
EricTRL, line 59: There's a mismatch between the default behaviour of this Mixin and the `@membership_required`-decorator.

docs/administrative/readme.md
```diff
@@ -0,0 +1,17 @@
+# Readme
+
+In this folder all information for the functionality of the Knights can be found.
+For the purpose of these documents, please adhere to the following definitions:
+
+**Squire**
+The internal platform the Knights as is active on the site
+
+**admins** 
+Persons with a high form of administrative power in the backend. In rule this should be board and the
```
EricTRL, line 10: I disagree that the board should be admins. They should have more rights than normal members (or else they can't actually do their administration), but they should not be able to touch certain settings (such as global permissions). Compare this to, say, the mailserver, where the board also does not get admin powers.

EricTRL (changes requested):
Changes seem good overall! I'm a bit worried about confusion for the board between `Member`, `Membership`, `MemberYear`, and `MemberLog` since they all have similar names. I don't really see an easy way to avoid that though.

I don't see a reason to keep `Member.has_paid_membership_fee` at this point though, given that this can now be properly indicated per year instead of having to reset it in September again and again.

I also like the confetti 🎉 (we really need more easter eggs)

membership_file/admin.py
```diff
@@ -24,6 +24,16 @@ def reset_has_paid_membership_fee(modeladmin, request, queryset):
             member.save()
 reset_has_paid_membership_fee.short_description = 'Reset membership fee paid status'
 
+
+def mark_as_current_member(modeladmin, request, queryset):
```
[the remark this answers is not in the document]
DutcherNL: Oops, fixed it. And written testcase and a bit more faillure catching

membership_file/util.py
DutcherNL: In  what way? The only thing that is different is that in the Mixin  one can disable the check for a member consideration. Which is by default taken into account just  like the decorator?

EricTRL: Seems like I read some part of the code wrong! You're absolutely right!

docs/administrative/readme.md
EricTRL: I agree that the board will pretty much always have access to (part of) the admin panel. Perhaps explicitly mention that the board won't have access to the entire backend?

EricTRL (approved):

PR #225, opened 2021-12-21
Upcoming Core Activities API

EricTRL:
On our website [kotkt.nl](https://kotkt.nl/) there is no clear overview of which activities the Knights organise. While there is a single link towards Squire's calendar page hidden in one of the sidebars, this is difficult to find and redirects to an entirely different platform. It would be useful for potential members to see which (core) activities are coming up.

To do so, we'll need an API. However, the only one that's currently available (and required for FullCalendar) finds all activities within a specific datetime-range. Additionally, it includes _all_ activities, even those that are not interesting for potential new members (E.g. a GMM, Christmas Dinner, etc.) or don't properly show what we as an association stand for (E.g. Videogame Thursday). These activities are not what are called "core activities".

This PR allows linking each activity to a specific `CoreActivityGroup`, which is a collection of activities from the same "pillar" of the association. E.g. boardgame days, roleplay-activities, or LARP-things. The API exposes the first (non-cancelled, non-removed) occurrence of a _single_ activity in each `CoreActivityGroup`. That is, if there are two activities in a single group, then only the first occurrence of one activity is returned through the API.


activity_calendar/api.py
```diff
@@ -69,3 +73,53 @@ def fullcalendar_feed(request):
...
+
+    # Add placeholders to ensure the order of the groupings does not change
+    earliest_moments = {
+        x:None for x in grouping_identifiers
+    }
+
+    # Iterate over all published activities that are part of the core groupings in the GET request
+    for activity in Activity.objects.select_related('core_grouping').filter(published_date__lte=now, core_grouping__identifier__in=grouping_identifiers):
+        # Fetch the next activitymoment (if any) for this activity that is not cancelled
+        activity_moment = activity.get_next_activitymoment(dtstart=now, exclude_cancelled=True)
```
DutcherNL, line 98: To futureproof this a bit more (just in case the default of get_next changes). Also, it looks nicer as you already differ from exclude_cancelled default.
```suggestion
        activity_moment = activity.get_next_activitymoment(dtstart=now, exclude_cancelled=True, exclude_removed=True)
```

activity_calendar/models.py
```diff
@@ -32,6 +32,17 @@
 def now_rounded():
     return timezone.now().replace(minute=0, second=0)
 
+class CoreActivityGrouping(models.Model):
+    """
+        A method to group certain core activities together. E.g. all boardgame evenings,
+        all roleplay-related actitvities, LARP-stuff, etc.
+    """
+    identifier = models.CharField(max_length=127, unique=True)
```
DutcherNL, line 40: It may be useful here to prevent empty strings here. Maybe add a `MinLengthValidator`? (https://docs.djangoproject.com/en/3.1/ref/validators/#minlengthvalidator)

DutcherNL (changes requested):

activity_calendar/api.py
```diff
@@ -69,3 +73,53 @@ def fullcalendar_feed(request):
...
+        if earliest_moment is None \
+                or earliest_moment.start_date > activity_moment.start_date:
+            earliest_moments[activity.core_grouping.identifier] = activity_moment
+
+    # Fetch data for the selected activitymoments
+    activity_moment_jsons = []
+    for activity_moment in earliest_moments.values():
+        if activity_moment is None:
+            continue
+
```
DutcherNL, line 110: Hmm, I think I accidentally lost this remark when submitting the review  again. So I'll type it again:
Whereas earlier you try to maintain order of the grouping identifiers, the order is completely rejected at this point. Particularly when an activitymoment was not found it just skips a step in the list. The json data also does not communicate the relevant grouping  identifiers. Could you rewrite the code to either just receive one identifier or, when allowing multiple to be added, have the result contain clarity on which activitymoment relates to which identifier.

activity_calendar/models.py
EricTRL: The field is required, so empty strings are already disallowed.

activity_calendar/api.py
EricTRL: The skipping was done on purpose (as activities might have been returned for other groupings in the same request). Added the grouping identifier in the returned activitymoment data 👍

DutcherNL (approved):

Issue #227, opened 2021-12-24
ActivitySlot clean method does not take ActivityMoments outside recurrence scheme into account


EricTRL:
`ActivitySlot`'s `clean` method prevents saving slots for activitymoments that fall outside their parent's recurrence scheme.

PR #228, opened 2021-12-24
Bugfix/activity slot check

DutcherNL:
Rewrote has_occurrence_at to get_occurrence_at to combine ActivityMoment retrieval to single point (simpling view)
Also addressed #227


activity_calendar/models.py
```diff
@@ -332,15 +332,40 @@ def duration(self):
...
+        :param date: Datetime instance of the occurrence
+        :param is_dst_aware: Whether the datetime instance is dst aware (relevant for recurrent activities)
+        :return:
+        """
+        # Check activitymoments on server
+        activitymoments = self.activitymoment_set.filter(recurrence_id=date)
+        if activitymoments.count() == 1:
+            return activitymoments.first()
+        elif activitymoments.count() > 1:
+            raise ActivityMoment.MultipleObjectsReturned()
```
EricTRL, line 348: Is this even possible? Each recurrence_id, parent_activity pair is unique.

activity_calendar/models.py
```diff
@@ -332,15 +332,40 @@ def duration(self):
...
+
+        # Find recurrent moments that have not yet been created on the server
+        if self.is_recurring:
```
EricTRL, line 351: (placed here for lack of a better location)
I think we should rewrite `Activity.is_recurring` to also return `True` if there also are activitymoments that don't match the activity's start date. This allows such activities/moments to be handled in the same way as RDATEs, which currently do make an Activity considered to be "recurring".

activity_calendar/tests/test_models.py
```diff
@@ -698,6 +698,43 @@ def test_get_next_activitymoment_cancelled(self):
...
+        occurrence = self.activity.get_occurrence_at(test_date)
+        self.assertIsNone(occurrence)
+        # Set time to 15:00 instead of 14:00 due to timeshift
+        test_date = timezone.datetime(2020, 11, 11, 15, 0, 0, tzinfo=timezone.utc)
+        occurrence = self.activity.get_occurrence_at(test_date)
+        self.assertEqual(occurrence.parent_activity, self.activity)
+        self.assertEqual(occurrence.recurrence_id, test_date)
+        # recurrence_id should be the date checked and not the date matching the recurrency pattern (i.e. 14:00)
+
+    def test_get_occurrence_at_db_activitymoment(self):
```
EricTRL, line 723: Duplicate testcase of the one below it? (if not, they still have the same method name)

activity_calendar/views.py
```diff
@@ -91,19 +91,10 @@ def setup(self, request, *args, **kwargs):
         self.activity = get_object_or_404(Activity, id=self.kwargs.get('activity_id'))
         self.recurrence_id = self.kwargs.get('recurrence_id', None)
 
-        self.activity_moment = ActivityMoment.objects.filter(
-            parent_activity=self.activity,
-            recurrence_id=self.recurrence_id
-        ).first()
+        self.activity_moment = self.activity.get_occurrence_at(self.recurrence_id)
```
EricTRL, line 94: NIce change; it moves the logic to the model instead of the view 👍

EricTRL (changes requested):
Nice changes 👍

activity_calendar/models.py
DutcherNL: Nope, it shouldn't. Hadn't realised that.

activity_calendar/models.py
DutcherNL: The term is_recurring is currently used as signifier for recurrce patterns only. Rewriting this function might introduce new complexities that I don't want to deal with at the moment. I agree with the remark though :)

EricTRL (approved):

PR #233, opened 2021-12-29
Update/catalogue tabs

DutcherNL:
Added tabs to the catalogue and an instructions page on how to use the catalogue


inventory/templates/inventory/inventory_instructions.html
```diff
@@ -0,0 +1,48 @@
...
+    </p>
+    <h3>Maintaining catalgoue</h3>
+    <p>
+        Each committee or member can see and edit their own items as registered in the sytem. Besides that some
+        committees have rights to maintain certain catalogues. For instance the BG (Boardgame committee) can maintain
+        the entire boardgame catalogue. This allows them to adjust the ownerships of any committee or member
+        on the boardgames. Mark certain items as no longer at the Knights or tweak notes on ownerships.
+    </p>
+    <p>
+        To add an item to the catalogue, press the `New &lt;Item type&rt;' button at the top of the catalogue screen.
```
EricTRL:
```suggestion
        To add an item to the catalogue, press the `New &lt;Item type&gt;' button at the top of the catalogue screen.
```

inventory/views.py
```diff
@@ -66,9 +66,31 @@ def get_context_data(self, *args, **kwargs):
...
+        """
+        Returns a list of dictionary objects for tab information
+        :return:
+        """
+        tabs = []
+        for item_type in Item.get_item_contenttypes():
+            if self.request.user.has_perm(f'{item_type.app_label}.view_{item_type.model}'):
+                model_class = item_type.model_class()
+                tabs.append({
+                    'verbose': model_class.__name__,
```
EricTRL, line 83: Why not use the class's verbose_name instead?

Also, you can consider adding the instructions page to the tabs directly instead of having a link to it on each page.

inventory/views.py
```diff
@@ -66,9 +66,31 @@ def get_context_data(self, *args, **kwargs):
...
+                    'icon_class': model_class.icon_class,
+                    'url': reverse("inventory:catalogue", kwargs={'type_id': item_type}),
+                    'selected': item_type == self.item_type,
+                })
+        return tabs
+
+
+class CatalogueInstructionsView(MembershipRequiredMixin, CatalogueMixin, TemplateView):
```
EricTRL, line 91: Ideally this should be limited to those with `boardgames.view_boardgames` (and friends) but this'll work for now

EricTRL (changes requested):

inventory/views.py
DutcherNL: Don't know why I didn't consider this simple but useful thing XD

inventory/views.py
DutcherNL: True, but didn't want to make the hassle. However, thinking about it now adding it in a later PR may be even more of a (if not forgotton) hassle

EricTRL (dismissed):

Ndoll1 (changes requested):
Could you also change "Catalogue for [x]" to be consistent with the rest (capitals etc)

Ndoll1 (approved):

PR #247, opened 2022-01-18
Feature/surveys

DutcherNL:
I've been toying with this idea for a while and sitting in the train allowed me to start it.

Pipeline

- [x] Surveys that can be answered by members
- [ ] Automatic retrieval for specific personal data (name, email) where required
- [x] Survey link to associationgroups for data access
- [x] Data visualisation in associationgroup page
- [ ] Download results as CSV (admin)
- [ ] Download results as CSV (associationgroup)
- [ ] Build survey (associationgroup pages)
- [ ] Link survey to activity (e.g. weekend)
- [ ] Announcement message for active unanswered survey (if applicable)
- [x] Survey activation period/window
- [ ] Documentation


surveys/question_types.py
```diff
@@ -0,0 +1,137 @@
...
+    def get_extra_form_field_kwargs(self, question, answer=None):
+        kwargs = super(IntQuestion, self).get_extra_form_field_kwargs(question, answer=answer)
+        if question.option_1:
+            kwargs['min_value'] = int(question.option_1)
+        if question.option_2:
+            kwargs['max_value'] = int(question.option_2)
+        return kwargs
+
+
+class ChoiceQuestion(QuestionType):
```
EricTRL, line 94: A multi-choice question (with checkboxes) could be a worthy addition.

surveys/views.py
```diff
@@ -0,0 +1,75 @@
...
+    pk_url_kwarg = 'survey_id'
+    context_object_name = 'survey'
+
+    def setup(self, request, *args, **kwargs):
+        super(SurveyFormView, self).setup(request, *args, **kwargs)
+        self.object = self.get_object()
+
+    def dispatch(self, request, *args, **kwargs):
+        if self.object.end_date is not None and self.object.end_date < now():
+            return self.render_expired_response()
```
EricTRL, line 34: Could pass on the request to a different View (inheriting from TemplateResponse) instead.

EricTRL (commented):
Code logic seems good. There's a bit to do regarding template layout and communication to the user, but I figured the current ones are still somewhat of a placeholder.

Another thing to think about is how to handle surveys with large amounts of options. For things like association merchandise, it's fine to have one question per merch item (E.g. T-shirt S, T-shirt M, T-shirt L, oven mittens, etc.).
However, for things like the (somewhat annual) boardgame bulk buy with its hundreds of boardgames that can be ordered, this approach isn't really feasible. Of course, this isn't necessarily something that we need to support initially, or even at all.

surveys/models.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+
+class Survey(models.Model):
+    name = models.CharField(max_length=32)
+    description = models.CharField(max_length=128)
+    created_on = models.DateTimeField(auto_now_add=True)
+    organisers = models.ManyToManyField(AssociationGroup, blank=True)
+    start_date = models.DateTimeField(blank=True, null=True)
+    end_date = models.DateTimeField(blank=True, null=True)
```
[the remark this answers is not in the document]
DutcherNL: I did aim for consisency, which was why I named it the same we do acivities XD

surveys/question_types.py
DutcherNL: I am intending to alter type registration a bit more, maybe in a too elaborate way, where each type will be represented in a QuestionType table. This way we can introduce new forms in the future if neccessary

surveys/templates/surveys/survey_overview.html
```diff
@@ -0,0 +1,36 @@
+{% extends 'core/base.html' %}
```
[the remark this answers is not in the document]
DutcherNL: This page is will probably be omitted in end result in favour of more localised pages on committee/ user page

PR #255, opened 2022-02-18
Implemented shortcut system in the app

DutcherNL:
Implemented new feature that allows shortcuts to be added if so desired:

e.g. app.kotkt.nl/gala

These can redirect to any url, so also external surveys and such.


core/migrations/0006_shortcut.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Shortcut',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('title', models.CharField(max_length=16)),
+                ('description', models.CharField(blank=True, max_length=128, null=True)),
+                ('location', models.CharField(max_length=32, unique=True)),
+                ('referrence_url', models.URLField()),
```
helaan, line 20: reference

core/templates/core/redirect_url_page.html
```diff
@@ -0,0 +1,26 @@
...
+<!DOCTYPE html>
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+    <!-- We imitate a normal page meta data so normal data can be obtained for link previews -->
+    <meta property="og:title" content="{{ shortcut.title|default:"Squire" }}" />
+    <meta property="og:description"
+          content="{{ shortcut.title|default:'Squire is the new webapp developed by the UUPS-committee of ESRG Knights of the Kitchen Table.'}}" />
```
helaan, line 12: Squire no longer new at this point, is it possible to not hardcode this anymore?

squire/urls.py
```diff
@@ -45,6 +45,9 @@
     path('', include('core.urls', namespace='core')),
 
     path('', include('user_interaction.urls')),
+
+    # Shortcuts, always at the end
```
helaan, line 49: maybe clarify why it should be at the end? (i guess shortcuts could overtake existing pages otherwise?)

core/models.py
```diff
@@ -84,3 +84,22 @@ def clean(self):
...
+    # A description that will appear when link is loaded through Open Graph previews
+    description = models.CharField(max_length=128, blank=True, null=True)
+
+    # The local name e.g. 'intro' can be found under <squire-domain>.nl/intro
+    location = models.CharField(max_length=32, unique=True)
+    # The url it references to. It should be the full url e.g. 'https://www.google.com'
+    referrence_url = models.URLField()
+
+    def __str__(self):
+        return f'{self.title} @ {self.location}'
```
helaan, line 105: nit: maybe use `->` instead of `@`, as the shortcut points to a given location?

core/shortcut_urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path, re_path
+
+from core.views import UrlRedirectView
+
+
+urlpatterns = [
+
+    re_path(r'^(?P<url_shortener>.*)/$', UrlRedirectView.as_view()),
```
helaan, line 8: Do you need a terminating slash to trigger the redirect? Preferably it'd work with both

helaan (commented):
took a quick glance through the code, didn't run it

core/templates/core/redirect_url_page.html
DutcherNL: Created new issue for this as it's litterally the current default text as extracted from base.html and thus outside the scope of this PR

core/shortcut_urls.py
DutcherNL: Squire (/Django) always automatically adds a / if one is not present in the requested url, so it's best to assume it will always end with a /

DutcherNL: If you are referring to the $ at the end, it's best to not catch any extra material after the url, and this way we can be sure that is not the case.

helaan (approved):

Issue #207, opened 2021-12-05
Internal information pages


DutcherNL:
In order to improve knowledge within the association. Squire could have use with some page(s) that display general information for members including, but not limited to:

- Statuen, HR and huisregel files
- Lords alumni information page (#206)
- Instructions about Scala Dining
- Communication channels (Telegram, Discord)

The Newsletter can also be moved in this branch.

DutcherNL:
Added Good first issue as the creation of the loose pages (i.e. Scala Dining instructions page) is a good first issue.

Ideally this can be combined through viewcollectives (see utils module). Setting this up should also be simple when you review how its done for e.g. the account system (located in user_interaction)

Ndoll1:
Membership card can be added as well on a later date. Right now it is located at Membership information

PR #262, opened 2022-03-02
Adding Membership Card to Membership page

Ndoll1:
Title is self explanatory


EricTRL (approved):
Can later be integrated in #207, but this works for now.

PR #266, opened 2022-03-20
Feature/calendars

DutcherNL:
Introduce specific calandars that activities can be connected to specific activities (not moments)

Also added birthday calendar feed


activity_calendar/feeds.py
```diff
@@ -253,3 +283,88 @@ def item_extra_kwargs(self, item):
...
+        # Declare a feed name
+        activity.feed_name = f"bday-{slugify(member.get_full_name())}"
+        return activity
+
+    def items(self):
+        """ Constructs a list of activities that represent the birthdays of the members """
+        # Create iterator that builds activities from the list of members
+        return map(
+            self.construct_birthday,
+            Member.objects.filter(
```
EricTRL, line 361: `Member.date_of_birth` can technically be `None` as it's not a required field

activity_calendar/migrations/0018_auto_20220312_2108.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 2.2.15 on 2022-03-12 20:08
```
EricTRL, line 1: Squash please :)

activity_calendar/admin.py
```diff
@@ -111,3 +113,5 @@ def recurrence_id_with_day(self, obj):
     inlines = [ParticipantInline]
 
 admin.site.register(ActivitySlot, ActivitySlotAdmin)
+admin.site.register(Calendar)
```
EricTRL, line 116: Adding an `Inline` here for `CalendarActivityLink` makes it way easier to manage this in the admin panel.

activity_calendar/urls.py
```diff
@@ -22,7 +22,9 @@
         ])),
     ])),
 
-    path('api/calendar/ical', CESTEventFeed(), name='icalendar'),
+    path('api/calendar/ical', PublicCalendarFeed(), name='icalendar'),
+    path('api/calendar/birthdays/', BirthdayCalendarFeed(), name='ical_birthdays'),
```
EricTRL, line 26: It should probably be noted in the calendar preferences form that the birthday calendar is public. In the future, we might want to generate a unique link per member.

activity_calendar/models.py
```diff
@@ -79,10 +80,12 @@ class Meta:
 
     # The date at which the activity will become visible for all users
     published_date = models.DateTimeField(default=now_rounded)
+    is_public = models.BooleanField(default=True, help_text="If activity should be on public calendar")
```
EricTRL, line 83:
```suggestion
    is_public = models.BooleanField(default=True, help_text="If activity should be on the public calendar (outside of Squire)")
```

activity_calendar/feeds.py
```diff
@@ -32,7 +37,14 @@ def func_wrapper(self, item):
 def get_feed_id(item):
     # ID should be _globally_ unique
     if isinstance(item, Activity):
-        return f"local_activity-id-{item.id}@kotkt.nl"
+        # Sometimes a specific feed_name is given along with the object. Prioritise that.
```
EricTRL, line 40: I'm not entirely convinced of this setup, but I don't really see a good alternative on a quick glance either.

activity_calendar/feeds.py
```diff
@@ -32,7 +37,14 @@ def func_wrapper(self, item):
...
+        if hasattr(item, 'feed_name'):
+            return f"local_activity-name-{item.feed_name}@kotkt.nl"
```
EricTRL, line 42:
```suggestion
            return f"{item.feed_name}@kotkt.nl"
```

EricTRL (changes requested):

activity_calendar/models.py
DutcherNL: Should also apply inside of Squire. Updated that now.

EricTRL (approved):

PR #273, opened 2022-04-11
Refactor/link activityslot to moment

### LINKED ISSUE #83: Link ActivitySlots to an ActivityMoment instead of an Activity (with a recurrence_id)
_Edit: Initial statement refers to recurrence_id in activityslots_

We are not using it in the front view,
We are not using it in the admin view
We are not using it in activities
We are not using it in permissities

It serves no use but cluttering
It doesn't do a single thing

I don't know about you lot
But I'd like it gone from every slot

EricTRL:
Modifies `ActivitySlot` to link to an `ActivityMoment` instead of an `Activity` (with a recurrence_id). When someone registers for a slotless activity or creates a slot for an activity that does not have associated ActivityMoment in the database yet, this is automatically created prior to the creation of the slot.

I've gone ahead and added Inlines for ActivityMoment to Activity (which could have already been there) as well as ActivitySlot to ActivityMoment (which is now actually possible). Also did some cleanup for some admin mixins.

Old slots are attached to an ActivityMoment through a database migration. If an activitymoment with a matching recurrence_id and parent_activity is not found for a slot, a new activitymoment is created with those parameters instead. There might still be some old ActivitySlot instances without a recurrence_id for non-recurring activities. The start_date of the matching activity is used as the recurrence_id instead.

Closes #83


EricTRL:
There's a small issue I've encountered: orphan activityslots. It seems that an activity with slots was moved in some cases, causing those slots to become "orphaned". For example, the Monopoly race was moved from March 15 to April 15. The old slots remained on March 15 and are invisible in Squire. Yet, during this migration, they obtain an automatically-created ActivityMoment and re-appear.

On the other hand, there are plenty of Slots which occur at some point in time for which there is no matching ActivityMoment.

There is no easy way to find out whether a slot is an orphan or not. We can include a copy of `ActivityMoment.is_part_of_recurrence` but for slots instead, but that doesn't really seem worth the hassle. I recon we should bite the bullet here and simply have some of those orphan slots appear, and manually get rid of the extra activitymoments they created. (E.g. the March 15 Monopoly race)

DutcherNL (approved):
I have honestly nothing to mark here. It's a useful addition for a cleaner code. Also impressed that you wrote testcases for your migration, though honestly, due to it's one-time use, I wouldn't have deemed that worth the investment.

DutcherNL (approved):

Issue #200, opened 2021-12-01
Refactor Achievements Module


EricTRL:
The achievements module does not use Class Based Views (CBVs), which the rest of the application does. Furthermore, this module has a dependency towards only a small part of the Django REST framework that is not really needed if CBVs would be properly utilised. Getting rid of this dependency would be nice.

PR #278, opened 2022-09-01
Refactor of Achievements

LenaWil:
I … eh … kinda forgot what I still needed to do…

I am soo bad at names, but I think it was Eric, who helped, could you help again?

#200


LenaWil:
Although, I can prolly take an other look myself first.

PR #281, opened 2022-09-10
Downloads share URL

EricTRL:
Added a share URL for association downloads, like statutes or other official protocols. The same logic already existed for the newsletter share URL, but was unified for the two of them. Note that this is only a temporary solution until we get around to #207 

The main navigation bar ordering was also changed to be ordered according to importance, rather than the date a tab was introduced.


core/templates/core/downloads.html
```diff
@@ -0,0 +1,27 @@
...
+{% block og-description %}
+    A central location for association protocols or documents.
+{% endblock og-description %}
+
+{% block content %}
+    <img src="{% static 'images/newsletter_header.png' %}" alt="Knights newsletter header image" width="100%" />
+
+    <h1>Downloads</h1>
+    <p>
+        Looking for official bureaucracy? All public protocols, decisions, and whatnot can be accessed <a href="{{ global_preferences.downloads__share_link }}">here</a>
```
LenaWil, line 22: Doesn’t end with a period.

core/tests/tests_frontend.py
```diff
@@ -44,47 +45,63 @@ def test_register(self):
...
+    def test_newsletter(self):
+        """ Tests if the newsletter global preference is working """
+        self._test_global_preference_page_404('newsletter__share_link', reverse('core:newsletters'), TestMemberUser)
+        self._test_global_preference_base_template_nav('newsletter__share_link', reverse('core:newsletters'), TestMemberUser)
+
+    def test_downloads(self):
+        """ Tests if the downloads global preference is working """
+        self._test_global_preference_page_404('downloads__share_link', reverse('core:downloads'), TestMemberUser)
+        self._test_global_preference_base_template_nav('downloads__share_link', reverse('core:downloads'), TestMemberUser)
+
```
LenaWil, line 104: I have to admit that I don’t know enough about the tests to say for sure if it covers everything, but it does pass and looks okay.

LenaWil (commented):
I will have to admit that I’m still not familiar enough with Django to say if the code style is good etc., but I see no bugs, except that one forgotten period.

LenaWil (dismissed):
I added a period.

core/tests/tests_frontend.py
EricTRL: CodeCov shows us which lines of changed code are not covered by tests directly within GitHub. For example, see CodeCov's comment in #247:

<details>

image
</details>

Of course, coverage doesn't tell us anything about the quality (or even correctness) of the tests, which is what these reviews are for. It seems like the diff doesn't really show well what actually changed here though, so let me walk you through it.

1. Although it shows up as changed, `def test_homepage_message(self):` wasn't actually touched at all.
2. For both the newsletter and downloads page, we want to test two things:
   1. Their individual pages `/newsletters` and `/downloads` should raise a HTTP 404 if no share link is set in the global preferences, and they should be accessible if such a share link is set.
   2. Links to these individual pages should not be present in the navigation bar if such a link is not set, and they should be linked to if it is set.
3. These two tests were already present for newsletters previously. I could have copy-pasted and adapted those tests for the Downloads and called it a day, but that would result in code duplication. Instead, I extracted two general methods based on the two newsletter tests, which get called with different parameters for Newsletters and Downloads.
`def _test_global_preference_page_404(..)` tests condition "i" mentioned above, whereas `def _test_global_preference_base_template_nav(..)` tests condition "ii". Tests were also moved to a different Test-class `NavGlobalPreferenceTest` for better organisation.

To verify that this change was made correctly, you could:
1. Check whether the old tests for newsletters were correctly generalised into the checks `def _test_global_preference_page_404(..)` and `def _test_global_preference_base_template(..)`
2. Check that the right parameters are used when calling these general methods.

core/templates/core/downloads.html
```diff
@@ -0,0 +1,27 @@
...
+
+{% block og-title %}
+    Knights Downloads
+{% endblock og-title %}
+{% block og-description %}
+    A central location for association protocols or documents.
+{% endblock og-description %}
+
+{% block content %}
+    <img src="{% static 'images/newsletter_header.png' %}" alt="Knights newsletter header image" width="100%" />
```
DutcherNL, line 18: As this now also used in downloads, rename the file to association_header.png or something like that

core/views.py
```diff
@@ -35,16 +35,27 @@ def logoutSuccess(request):
...
+class GlobalPreferenceRequiredMixin:
+    """ Mixin that requires a specific global_preference to be non-empty or True. Throws a HTTP 404 otherwise. """
+    global_preference = None
+
+    def setup(self, request: HttpRequest, *args, **kwargs) -> None:
+        super().setup(request, *args, **kwargs)
+
+        preference = global_preferences[self.global_preference]
+        if not preference:
+            raise Http404()
```
DutcherNL, line 48: I like the class based views rewrite

core/tests/tests_frontend.py
LenaWil: > Of course, coverage doesn't tell us anything about the quality (or even correctness) of the tests, which is what these reviews are for.

Yeah, I meant it that way.  
And I wasn’t familiar enough to assess that properly.  
I mean the tests looked they made sense, and I got most of what you said in your comment already, but I missed `page_url` thing @DutcherNL pointed out, which I agree on.

EricTRL: I agree the comments could have been a lot more helpful there. I've improved them in the latest commit 👍

DutcherNL (approved):

PR #286, opened 2022-10-04
Slot images may not show

EricTRL:
ActivitySlots without their own PresetImage do not show any image at all, rather than inheriting them from their parent activity(moment). Instance methods were renamed to be clearer of what they represent.

Also disables tests for an old data migration. There's no point in continuously running these.


DutcherNL (approved):
I assume thie is not the PR that Mick asked me to auto-approve

Issue #13, opened 2019-08-19
Add code linting


helaan:
Add Django-specific code linting to prevent syntax errors and to keep the styling consistent.

PR #290, opened 2023-01-07
Extended MembershipMiddleware to catch UserIsNotAMember exceptions

DutcherNL:
Adjusted membershipMixin to throw error, which is to be catched by the Membership Middleware. This prevents accidental overwrites of the response when a user is not a member. Also removes the ugly redirect when a user is not allowed on a page.


membership_file/util.py
```diff
@@ -58,5 +60,5 @@ def dispatch(self, request, *args, **kwargs):
             return HttpResponseRedirect(resolve_url(self.fail_url))
         if not request.member.is_considered_member() and self.requires_active_membership:
             # Current session has a disabled member connected
-            return HttpResponseRedirect(resolve_url(self.fail_url))
+            raise UserIsNotCurrentMember()
```
LenaWil, line 63: Why isn’t the first line changed, but the second one is?

membership_file/tests/test_middleware.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_process_request_anonymoususer(self):
+        request = RequestFactory().get('')
+        request.user = AnonymousUser()
+        self.middleware.process_request(request)
+        self.assertEqual(request.member, None)
+
+    def test_process_exception(self):
+        request = self._create_request(1)
+        self.assertIsNone(self.middleware.process_exception(request, Exception()))  # Ignore irrelevant errors
```
LenaWil, line 38: Which “irrelevant errors” are you ignoring?  
Or are you checking if there are no other errors?

membership_file/tests/test_middleware.py
```diff
@@ -0,0 +1,43 @@
...
+
+        response = self.middleware.process_exception(request, UserIsNotCurrentMember())
```
LenaWil, line 40: I don’t know that much about what should and shouldn’t be tested, but is maybe useful to add a second test to check if util.py’s `MembershipRequiredMixin.dispatch()` actually generates the exceptions? (And maybe a sample page that should throw this too, but that might be too much)

(Please feel free to disagree with this one though.)

membership_file/middleware.py
```diff
@@ -13,3 +14,14 @@ def process_request(self, request):
...
+    def process_exception(self, request, exception):
+        if isinstance(exception, UserIsNotCurrentMember):
+            return TemplateResponse(
+                request=request,
+                template='membership_file/no_member.html',
+                context={},
+                status=403,
+            )
+
+
```
LenaWil, line 27: I know this is on me for not doing #13, but why the new lines?

LenaWil (changes requested):
I have a couple question about the code, and also a very minor formatting issue.

membership_file/tests/test_middleware.py
DutcherNL: Here I want to intercept only UserIsNotAMember exceptions, any other exception should be ignored (eg HTTP404 exception). I want specific behaviour in case that exception is thrown (return that template). Check the Django Middleware documentation on what Middleware does if you want to know why and how this method is triggered https://docs.djangoproject.com/en/dev/topics/http/middleware/#process-exception

membership_file/tests/test_middleware.py
DutcherNL: I agree, I will do this later as I wrote testcases for the MembershipRequiredMixin in the nextcloud PR already and will update that based on this behaviour once merged

membership_file/middleware.py
DutcherNL: No conscious reason. Likely done so I had more visual clarity in my screen (blank space). I can remove them

membership_file/tests/test_middleware.py
LenaWil: > any other exception should be ignored

With ignored, you mean catch and ignore or just letting it be thrown?

membership_file/util.py
DutcherNL: Because you have eyes and I don't 😎

membership_file/tests/test_middleware.py
LenaWil: 👍

LenaWil (approved):

LenaWil (changes requested):
Wait, I forgot something.

membership_file/tests/test_middleware.py
LenaWil: Okay, would this be a correct comment?

```suggestion
        self.assertIsNone(self.middleware.process_exception(request, Exception()))  # Check if irrelevant errors are ignored
```

LenaWil (approved):
Sorry it took so long.

PR #291, opened 2023-01-14
Management/group control

DutcherNL:
With two PRs nearly ready that extends on the AssociationGroup CollectiveView I figured it might be nice to have a better control than remembering and searching for the correct permissions for the correct tabs. So I automated that a bit:
afbeelding


Ndoll1 (approved):
committees/admin/forms.py

Test file updaten, voor de rest prima. Keur het wel goed

PR #293, opened 2023-01-17
Update/versatile association group controls

DutcherNL:
In preparation to the  introduction of campaign groups, some changes have been made to the way that association_groups are presented. 
- There is now a home tab that can distinguish based on group type which form needs to be displayed
- Introduction of a settings tab where options can hook into depending on the association_group
- Created visual for campaigns shield


committees/migrations/0006_auto_20220417_1443.py
```diff
@@ -0,0 +1,50 @@
...
+def reverse_func(apps, schema_editor):
+    # forwards_func() creates two Country instances,
+    # so reverse_func() should delete them.
+    AssociationGroup = apps.get_model("committees", "AssociationGroup")
+    Group = apps.get_model("auth", "Group")
+    db_alias = schema_editor.connection.alias
+    for association_group in AssociationGroup.objects.using(db_alias).all():
+        if association_group.site_group is None:
+            group = Group.objects.alias(db_alias).create(
+                nmae = association_group.name,
```
EricTRL, line 25:
```suggestion
                name = association_group.name,
```

committees/migrations/0006_auto_20220417_1443.py
```diff
@@ -0,0 +1,50 @@
+# Generated by Django 2.2.15 on 2022-04-17 12:43
```
EricTRL, line 1: Rename file to indicate this is a data migration

committees/migrations/0009_merge_20230121_1237.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.2.16 on 2023-01-21 11:37
```
EricTRL, line 1: This migration is empty?

committees/templates/committees/committee_pages/group_detail_info.html
```diff
@@ -1,15 +1,11 @@
...
     container
 {% endblock %}
 
 
-{% block content %}
-{{ block.super }}
-
-
-
+{% block setting_option %}
```
EricTRL, line 8: There's weird empty space on the left.

committees/models.py
```diff
@@ -9,19 +9,26 @@
 
 
 class AssociationGroup(models.Model):
-    site_group = models.OneToOneField(Group, on_delete=models.CASCADE)
+    site_group = models.OneToOneField(Group, on_delete=models.CASCADE, blank=True, null=True)
+    name =  models.CharField(max_length=150)
```
EricTRL, line 13:
```suggestion
    name =  models.CharField(max_length=150, unique=True)
```

EricTRL (changes requested):
Looking good, just a few minor things!

committees/backends.py
```diff
@@ -21,6 +21,10 @@ def has_perm(self, user_obj, perm, obj=None):
             content_type__app_label=app_label,
             codename=codename,
             group__associationgroup__members__in=[member],
+        ).exists() or Permission.objects.filter(
```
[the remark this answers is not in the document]
DutcherNL: I understand what you mean about depracting the link, but I'll keep it like this to ensure there won't suddenly be a conflict on the site that I might have missed. Another use of this approach is that it's easier to move board permissions to a different group.
That said, I do prefer redacting it in the long run, but I think that should be a seperate PR as depracting anything should not be done in a massive PR like this (to prevent any issues unexpected issues arising in working branches)

committees/migrations/0009_merge_20230121_1237.py
DutcherNL: It was a merge migrations that merges two branches, I reordered some migration orders so this migration is no longer needed.

EricTRL (approved):

PR #304, opened 2023-02-14
Feature/meetings

DutcherNL:
A new feature to insert meetings into squire. This will be implemented for campaign planning

Still to do:
- [x] Redact is_public field (it is replaced with a choicefield)
- [x] Meeting calendar feed
- [x] Testcases


activity_calendar/constants.py
```diff
@@ -0,0 +1,20 @@
+##############################################################################
+##                                                                          ##
+##  This file contains constant values that are used throughout the module  ##
+##  To prevent circular loops these are defined here                        ##
+##                                                                          ##
+##############################################################################
+
+
+ACTIVITY_PUBLIC = "PUBLIC"
```
EricTRL, line 9: Consider making this an Enum?

EricTRL (changes requested):

EricTRL (approved):

LenaWil (changes requested):
Friendly request to at least run the formatter on the files you added

It’s possible to do that offline.

If you really don’t want to or think the formatter is too strict, that’s okay too, but then we should talk about improving that.

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

DutcherNL:
> Friendly request to at least run the formatter on the files you added
Is there a way to run them all as a single commit. As you can see it has a lot of suggestions and I rather not gain RSI from formatting xD

LenaWil:
> 
> Is there a way to run them all as a single commit. As you can see it has a lot of suggestions and I rather not gain RSI from formatting xD
XD

Which IDE are you using?

LenaWil (commented):
I sent #314, otherwise fine with me.

LenaWil (approved):

PR #305, opened 2023-02-15
Add TSV, ODS & XLSX to export options

### LINKED ISSUE #301: Support .tsv because excel is a terrible program that gets confused with comma
Originally reported by Denise.

If you export the members using .csv and import it to excel on a Dutch system, excel gets confused because numbers are divided by `,` and not `.` in the Netherlands.

Therefore I propose we support `.tsv` as export target too.

LenaWil:
TSV with BOM works fine, but Excel doesn’t open them by default. ODS doesn’t work properly if both unicode and Excel are used.
XLSX of course does work correctly, but I don’t want to use a microsoft specific format.

Fixes #301.

I should still write tests, but I would like others to review it first.

### DIFF
--- a/membership_file/admin.py
+++ b/membership_file/admin.py
@@ -1,6 +1,5 @@
 
 from datetime import datetime
-import unicodedata
 
 from django.contrib import admin, messages
 from import_export.admin import ExportActionMixin
--- a/membership_file/admin.py
+++ b/membership_file/admin.py
@@ -1,9 +1,10 @@
 
 from datetime import datetime
+import unicodedata
 
 from django.contrib import admin, messages
 from import_export.admin import ExportActionMixin
-from import_export.formats.base_formats import CSV
+from import_export.formats.base_formats import CSV, TSV, ODS, XLSX
 
 from .forms import AdminMemberForm
 from .models import Member, MemberLog, MemberLogField, Room, MemberYear, Membership
@@ -11,6 +12,11 @@ from core.admin import DisableModificationsAdminMixin, URLLinkInlineAdminMixin
 from membership_file.export import MemberResource, MembersFinancialResource
 from utils.forms import RequestUserToFormModelAdminMixin
 
+class TSVUnicodeBOM(TSV):
+    def get_title(self):
+        return "tsv with Unicode BOM marker"
+    def export_data(self, *args, **kwargs):
+        return unicodedata.lookup('ZERO WIDTH NO-BREAK SPACE') + super().export_data(*args, **kwargs)
 
 class HideRelatedNameAdmin(admin.ModelAdmin):
     class Media:
@@ -49,7 +55,7 @@ class MemberWithLog(RequestUserToFormModelAdminMixin, ExportActionMixin, HideRel
     ##############################
     #  Export functionality
     resource_class = MemberResource
-    formats = (CSV,)
+    formats = (CSV,XLSX,TSVUnicodeBOM,ODS,)
 
     def has_export_permission(self, request):
         return request.user.has_perm('membership_file.can_export_membership_file')
@@ -214,7 +220,7 @@ class MemberYearAdmin(ExportActionMixin, admin.ModelAdmin):
     ##############################
     #  Export functionality
     resource_class = MembersFinancialResource
-    formats = (CSV,)
+    formats = (CSV,TSVUnicodeBOM,ODS,XLSX,)
 
     def has_export_permission(self, request):
         return request.user.has_perm('membership_file.can_export_membership_file')
@@ -241,3 +247,4 @@ class MemberYearAdmin(ExportActionMixin, admin.ModelAdmin):
 class MembershipAdmin(admin.ModelAdmin):
     list_display = ['member', 'year', 'has_paid', 'payment_date']
     list_filter = ['year', 'has_paid']
+
--- a/membership_file/admin.py
+++ b/membership_file/admin.py
@@ -13,6 +13,8 @@ from membership_file.export import MemberResource, MembersFinancialResource
 from utils.forms import RequestUserToFormModelAdminMixin
 
 class TSVUnicodeBOM(TSV):
+    '''..tsv that starts with a `ZERO WIDTH NO-BREAK SPACE`, which is a Byte Order Marker, which forces Excel to recognise it as Unicode.
+    More info: https://en.wikipedia.org/w/index.php?title=Byte_order_mark&oldid=1135118973#Usage'''
     def get_title(self):
         return "tsv with Unicode BOM marker"
     def export_data(self, *args, **kwargs):
@@ -220,7 +222,7 @@ class MemberYearAdmin(ExportActionMixin, admin.ModelAdmin):
     ##############################
     #  Export functionality
     resource_class = MembersFinancialResource
-    formats = (CSV,TSVUnicodeBOM,ODS,XLSX,)
+    formats = (CSV,XLSX,TSVUnicodeBOM,ODS,)
 
     def has_export_permission(self, request):
         return request.user.has_perm('membership_file.can_export_membership_file')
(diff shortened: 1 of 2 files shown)


LenaWil:
Maybe I should also include a BOM in the .csv exports… what are y’alls thought on it?

membership_file/admin.py
```diff
@@ -1,16 +1,22 @@
...
 from .models import Member, MemberLog, MemberLogField, Room, MemberYear, Membership
 from core.admin import DisableModificationsAdminMixin, URLLinkInlineAdminMixin
 from membership_file.export import MemberResource, MembersFinancialResource
 from utils.forms import RequestUserToFormModelAdminMixin
 
+class TSVUnicodeBOM(TSV):
+    def get_title(self):
+        return "tsv with Unicode BOM marker"
+    def export_data(self, *args, **kwargs):
+        return unicodedata.lookup('ZERO WIDTH NO-BREAK SPACE') + super().export_data(*args, **kwargs)
```
DutcherNL, line 19: Why does this character need to be added in front of the TSV file? Can you leave a comment to explain that?

membership_file/admin.py
```diff
@@ -214,7 +220,7 @@ class MemberYearAdmin(ExportActionMixin, admin.ModelAdmin):
     ##############################
     #  Export functionality
     resource_class = MembersFinancialResource
-    formats = (CSV,)
+    formats = (CSV,TSVUnicodeBOM,ODS,XLSX,)
```
DutcherNL, line 223: I don't know if the admin orders these types, but in case not. Please keep the order the same as defined in the MemberWithLog admin on line 58

LenaWil: Oh, that was an oversight, sorry.

membership_file/admin.py
LenaWil: I added a docstring now. Is this enough?

DutcherNL (approved):

DutcherNL:
> Okay, how extensive should the tests be?

Depends on how important/ how easy it is to accidentally break it. In your case I would only add a test that ensures export_data indeed adds the 'zero width no break space' to the beginning of the file.

membership_file/tests/tests_export.py
```diff
@@ -31,14 +31,16 @@ def test_accessible_rooms(self):
...
+                              row["accessible_rooms"])
+                self.assertIn(str(Room.objects.get(id=2)),
+                              row["accessible_rooms"])
                 break
 
 
 class MembershipFinanceFileExportTest(TestCase):
     """ Tests the structure of the exported membership file """
-    fixtures = ['test_users', 'test_members',]
+    fixtures = ['test_users', 'test_members', ]
```
LenaWil, line 43: Unrelated, but are all the fixtures used btw?

DutcherNL: Many fixtures contain references to other fixtures. For instance, `test_members` call users from `test_users`. The same applies for many other fixtures. `test_activity_slots` also use `test_members`, or the fixture for the committees module which uses `test_membes` as well.

membership_file/admin.py
```diff
@@ -60,25 +69,30 @@ def get_export_filename(self, request, queryset, file_format):
...
         'memberyear',
         'is_deregistered', 'marked_for_deletion',
         'is_honorary_member',
         'educational_institution',
-        ('tue_card_number', admin.EmptyFieldListFilter), ('external_card_number', admin.EmptyFieldListFilter),
-        ('key_id', admin.EmptyFieldListFilter), ('phone_number', admin.EmptyFieldListFilter),
+        ('tue_card_number', admin.EmptyFieldListFilter), ('external_card_number',
+                                                          admin.EmptyFieldListFilter),
+        ('key_id', admin.EmptyFieldListFilter), ('phone_number',
+                                                 admin.EmptyFieldListFilter),
```
DutcherNL, line 91: It's neater to place the enter outside of the brackets, than inside in this case.

```suggestion
        ('tue_card_number', admin.EmptyFieldListFilter),
        ('external_card_number', admin.EmptyFieldListFilter),
        ('key_id', admin.EmptyFieldListFilter),
        ('phone_number', admin.EmptyFieldListFilter),
```

membership_file/admin.py
```diff
@@ -87,23 +101,24 @@ def get_export_filename(self, request, queryset, file_format):
...
+                             'external_card_deposit', 'accessible_rooms']}),
         ('Legal Information', {'fields':
-            ['educational_institution', 'student_number',
-            'date_of_birth', 'legal_name']}),
+                               ['educational_institution', 'student_number',
+                                'date_of_birth', 'legal_name']}),
         ('Notes', {'fields':
-            ['notes']}),
+                   ['notes']}),
     ]
```
DutcherNL, line 122: I get where you are going from with the way you place these indices. But personally I think this worsens readability due to its very busy nature. All items that you indented basically have the same function, 'defining availlable fields in a given section' (notice how each line 103, 107, 112, 117, and 120 all end with `'fields':` ). However, due to its variation in indent spaces this is no longer instantly noticeable. Arguably you could move the list brackets outside the selection, but other than that I'd rather see it stay the same.
```
('Legal Information', {'fields':[
   'educational_institution', 'student_number',
   'date_of_birth', 'legal_name'
]}),
```

I do really appreciate you wanting to add code linting and standards into the code. I just believe that this specific implementation here has a counter productive effect

DutcherNL (changes requested):

membership_file/admin.py
LenaWil: If I am honest, I did not think that deeply about it, just a habit of using the formatter at every commit and I think it might have been an accident. I am working on it in a different branch.

I am currently looking at what each formatter makes from it.

`black` thinks this is the best solution: 
<details><summary>Verbose python code</summary>

```python

    fieldsets = [
        (
            None,
            {
                "fields": [
                    "user",
                    ("first_name", "tussenvoegsel", "last_name"),
                    "marked_for_deletion",
                    ("last_updated_date", "last_updated_by"),
                ]
            },
        ),
        (
            "Membership Status",
            {"fields": ["is_deregistered", "is_honorary_member", "member_since"]},
        ),
        (
            "Contact Details",
            {
                "fields": [
                    "email",
                    "phone_number",
                    ("street", "house_number", "house_number_addition"),
                    ("postal_code", "city"),
                    "country",
                ]
            },
        ),
        (
            "Room Access",
            {
                "fields": [
                    "key_id",
                    "tue_card_number",
                    (
                        "external_card_number",
                        "external_card_digits",
                        "external_card_cluster",
                    ),
                    "external_card_deposit",
                    "accessible_rooms",
                ]
            },
        ),
        (
            "Legal Information",
            {
                "fields": [
                    "educational_institution",
                    "student_number",
                    "date_of_birth",
                    "legal_name",
                ]
            },
        ),
        ("Notes", {"fields": ["notes"]}),
    ]
```

</details>

It’s certainly verbose, but it does work… What’s your opinion on it? 

A little of topic and prolly fits better in #13, but I kinda want to go with `black`, ’cause it’s easy to plugin and doesn’t leave much room for error, but it’s very opinionated, and will prolly change a lot of our formatting, at least a lot more than autopep, which I used now, did. Are you fine with that?

LenaWil: Otherwise I can restore it to it original state, and I have to look further on formatters.

LenaWil: Other problem with black is that it’s not really consistent with fields being one or multiple lines. (For this specific piece of code)

LenaWil: Huh, I thought I changed this back.

LenaWil: It seems I did?
But then why is it still listed as a diff here?

@DutcherNL Can you check if it's correct?

DutcherNL (approved):

PR #314, opened 2023-04-08
Add exception linting of migrations and some urls.py, format others

LenaWil:


inventory/urls.py
```diff
@@ -8,7 +8,7 @@
 
 
 class CatalogueConverter:
-    regex = '[\w.-]+'
+    regex = r"[\w.-]+"
```
LenaWil, line 11: IDK, but something was complaining.

PR #333, opened 2023-09-18
Show private activities too on calender when logged in

LenaWil:
Otherwise they are so hidden that it is useless. I am still working on the pr, but what do y’all think?

(I couldn’t quickly find where to set the user in the test)

Todo:
- [x] fix tests that fail currently.
- [ ] write test logging out and in with activities.
- [ ] check layout webpage. maybe edit to make clear it works on Apple packages.


user_interaction/views.py
```diff
@@ -95,7 +95,14 @@ def get_context_data(self, **kwargs):
             welcome_name = optimise_naming_scheme(welcome_name)
 
         activities = []
-        for activity in Activity.objects.filter(published_date__lte=timezone.now(), type=ActivityType.ACTIVITY_PUBLIC):
+        for activity in (
+            Activity.objects.filter(published_date__lte=timezone.now())
+            .filter(
+                Q(type=ActivityType.ACTIVITY_PUBLIC)
+                | (Q(type=ActivityType.ACTIVITY_MEETING) & Q(organisers__members=self.request.member))
```
EricTRL, line 102: I'm not sure what will happen if `self.request.member = None`. Might be useful to check that condition first just to be safe. Something along the lines of (pseudocode):
```python
activities = Activity.objects.filter(PUBLIC-Q)
if member is not None:
    activities = activities.filter(MEETINGS-Q)
for activity in activities:
    ...
```

EricTRL (changes requested):
I agree with the changes, but tests should be added.

Each `TestCase` class has a `client` attribute. You can set a user there through `self.client.force_login(<user>)`. If you're working with raw requests, you can set the `Request.user` attribute directly as well.

LenaWil:
> You can set a user there through `self.client.force_login(<user>)`.

How do you log out an user, `self.client.force_login(None)`?

.github/workflows/run_tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       DJANGO_ENV: "TESTING"
     strategy:
       matrix:
-        python: ["3.8"]
+        python: ["3.10"]
```
EricTRL: I think we're still running Python 3.8 in production, so please keep the two identical. Most type hints are already available in 3.8. E.g. `List[str]` (with an import for `List` from Python's `typing` module) opposed to `list[str]`

LenaWil: Yeah I already reverted this before you commented, but thanks for the feedback!

PR #334, opened 2023-09-22
Registration emails

EricTRL:
Adds a registration form in the Member admin panel that allows sending a registration email to the registered member. This email allows the registered member to create (or log in to) a Squire account in order to link it to the registered membership data. This form also allows the admin to immediately add rooms or add a membership for an active year.

This works identical to Django's password reset emails. A token is generated and attached to a URL in an email. When visited, the URL token is verified and stored in the session data to prevent HTTP Referer leaks. Further views retrieve the token from the session data (and verify it again). Tokens expire in 3 days.

When a user that is not logged in visits the URL, they are redirected to a registration form with the option to also log in with an existing account. If a user is already logged in and visits the URL, they are redirected to a login form with no option for a new account registration. If the logged in user is already a member, or if (another) user was already linked to the member, the URL won't work.

A few utilities are introduced as well:
- `OtherSelectWidget`. A radio button-based widget that also includes an "other" option for a free text entry. Useful in case some options are very common (e.g. country, educational institution)
- `ModelAdminFormViewMixin` and `FieldsetAdminFormMixin`. A view-form pair that allows using some of Django `ModelAdmin`'s features, notably fieldsets or default widgets (e.g. admin datetime picker). Useful for additional forms on the admin panel.

New dependencies:
- [DEV] aiosmtpd: Python's smtp module is deprecated.
- [ALL] django_object_actions: Extra action buttons in the admin panel.

Future work (not this PR):
- Email footer: The registration email footer is a global preference. This can be refactored once #317 is merged.


core/fixtures/test_users.json
Ndoll1, line 1: Is er een reden dat je deze mee stuurt?

core/static/images/bankpas.webp
Ndoll1, line 1: Zou deze afbeelding niet gebruiken. Met een snelle blik lijken veel gegevens nog erop te staan die enigzins nog relevant zijn.

Ndoll1 (approved):

core/static/images/bankpas.webp
EricTRL: Gegevens zijn uitgeblurred en dus niet leesbaar.

core/fixtures/test_users.json
EricTRL: Deze worden gebruikt in testcases. Gegevens zijn uiteraard niet geldig in productie

PR #348, opened 2024-09-07
Room access requirements + card updates

EricTRL:
- Remove `external_card_cluster` (not used anymore)
- All card numbers require 8 digits now (all 7-length numbers are prepended with a 0)
- External card digits require 4 numbers (all existing 3-digit numbers are prepended with a 0)
- Rooms now have an access type (card, key, other)
- When registering a member, Squire no longer requires a phone number if 1) only rooms with card access are requested and 2) a tue card number is provided
- Updated Rickrolls to a themed variant


user_interaction/views.py
Ndoll1, line 1: Helaas kan ik deze niet goedkeuren. Fortnite monopoly is niet meer dan een stuk karton.

Ndoll1 (changes requested):

Ndoll1 (approved):

PR #359, opened 2025-09-01
Update welcome email

ThatSnowman:
changed the registration email to the new mail provided by the new board

PR #365, opened 2025-10-09
Welcome mail hotfix

LenaWil:
A hotfix for the mail so the message is accurate.

Not sure if there was something else that was broken.

### DIFF
--- a/membership_file/templates/membership_file/registration/registration_email.txt
+++ b/membership_file/templates/membership_file/registration/registration_email.txt
@@ -61,8 +61,7 @@ You can register an account for {{ member.get_full_name }} through the
 following link. Doing so will automatically link it to your membership
 data.</p></li>
 </ul>
-<p>{{ protocol }}://{{ domain }}{% url ‘membership:link_account/confirm’
-uidb64=uid token=token %}</p>
+<p>{{ protocol }}://{{ domain }}{% url 'membership:link_account/confirm' uidb64=uid token=token %}</p>
 <p>You can browse upcoming events, access the full list of our board
 games, and sign up for certain activities.</p>
 <ul>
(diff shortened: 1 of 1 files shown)


membership_file/templates/membership_file/registration/registration_email.txt
```diff
@@ -1,17 +1,80 @@
...
+data.</p></li>
+</ul>
+<p>{{ protocol }}://{{ domain }}{% url 'membership:link_account/confirm' uidb64=uid token=token %}</p>
+<p>You can browse upcoming events, access the full list of our board
+games, and sign up for certain activities.</p>
+<ul>
+<li><strong>WhatsApp / Telegram Announcement Channel –</strong> Join one
+of our announcement channels to make sure you never miss an event!<br />
+<a
+href="https://chat.whatsapp.com/G47fmdu3Ufh5rpBI9XQY7v?mode=ems_copy_c">WhatsApp</a><br />
```
EricTRL, line 71: These hardcoded links are a no-go. The Scala Whatsapp chat has an active botting problem and needs to cycle its links every once in a while. Check out how `sender.extra_description` is set up for inspiration on setting up a global preference.

EricTRL (changes requested):
Adding HTML tags in the template itself won't work straight out of the box, as it's still sent as plaintext. Check out where this file is called from (Ctrl+F), there should already be support to include an HTML-template, though that should still be activated.

membership_file/templates/membership_file/registration/registration_email.txt
LenaWil: oh in that case #359 is prolly better to do direct and I will go help @ThatSnowman there.

EricTRL:
This isn't needed anymore; these changes were included in the previous base branch.

PR #371, opened 2026-01-06
Better Nextcloud integration error reporting

LenaWil:
The first commit is the quick change I did to have better error reporting.


squire/settings.py
```diff
@@ -26,14 +26,18 @@
 DEBUG = os.getenv("DJANGO_ENV") != "PRODUCTION"
 
 if os.getenv("SENTRY_DSN"):  # pragma: no cover
+    print("Sentry trying to initialize...")
```
LenaWil: I need to remove this, and also maybe the extra sentry stuff.

Ndoll1 (approved):

PR #372, opened 2026-02-19
upgrade pillow because of CVE

LenaWil:


LenaWil:
I have finally tested this and am approving this pull request since it has an CVE.

PR #373, opened 2026-02-24
Show cancelled events as cancelled in full calendar overview

### LINKED ISSUE #369: Cancelled events still show up in calender
If you cancel an event it still shows up in the month view of activities. Only when you click on the event you can see that it is cancelled.

LenaWil:
Fixes #369.

If anyone can think of a better place to add the cancelled text, let me know.

I still haven’t found the duplicate calendar bug that seems to also happen, and fixing the bug as described was relatively simple.

Todo:
- [ ] ~~write test~~
- [x] look into other iCal bug.


LenaWil:
Okay I can’t easily reproduce the other part of the bug, so will look into that later.

LenaWil:
Also since this is mostly a visual bug I don’t think it needs a test.

PR #374, opened 2026-02-24
Upgrade Django to version 5.2

LenaWil:
Seems to mostly works, except for one test. 

Also 4 will stop getting supported in a while.

6 doesn’t, because of `path("<assoc_group:group_id>/", registry.get_urls(with_namespace=False)),` complaining about a duplicated assoc_group registration.


core/admin_status/views.py
```diff
@@ -15,7 +15,7 @@ class LogFileView(AdminStatusViewMixin, TemplateView):
     """
 
     tags = {
-        "&lt;[a-z0-9_\-:\s]*&gt;": "text-danger",
+        r"&lt;[a-z0-9_\-:\s]*&gt;": "text-danger",
```
LenaWil, line 18: @EricTRL I got a `SyntaxWarning: "\-" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\-"? A raw string is also an option.` on a modern Python version and it since the escapes seemed more for the regex, than for python, and changing it into a raw string didn’t seem to change behaviour on both the older and newer python, I did that, but just in case, can you double check this?

EricTRL: An r-string should be fine here yeah. This is regex syntax that's included later down below. This essentially matches any letter, digit, _, -, :, and space between `<` and `>`

EricTRL (approved):
Seems fine. No breaking changes according to the 5.0-5.2 upgrade guides

PR #379, opened 2026-09-15
Update extend_membership_successpage.html

ofdekoning:
Added more info to what needs to be in the disciption of the bank account


LenaWil (approved):

membership_file/templates/membership_file/extend_membership_successpage.html
```diff
@@ -24,6 +24,9 @@ <h1>The quest goes on!</h1>
     <p><b>
         NL49 RABO 0325427771
     </b></p>
+	<p>
+		With is the description: Contribution [First Name][Last Name][Academic Year]
```
LenaWil: I read over this the first time, but „With _is_ the description”?

LenaWil (changes requested):
Actually I take it back.

PR #380, opened 2026-09-15
Add more instructions about extending membership

Kat-zorn:
See email from Jori.


membership_file/templates/membership_file/extend_membership_successpage.html
```diff
@@ -24,6 +24,9 @@ <h1>The quest goes on!</h1>
     <p><b>
         NL49 RABO 0325427771
     </b></p>
+    <p>
+        In the name of <b>“Eindhovens Studenten Rollenspel Genootschap Knights of the Kitchen Table”</b>, and with the payment description <b>“Contribution [first name] [last name] 2026-2027”</b>.
```
LenaWil:
```suggestion
        In the name of <b>“Eindhovens Studenten Rollenspel Genootschap Knights of the Kitchen Table”</b>, and with the payment description <b>“Contribution [first name] [last name] [Academic Year]”</b>. 
```
The member year changes every year, don’t hard code this. It is in theory possible to read this from Squire.

LenaWil (changes requested):
The year will change every

membership_file/templates/membership_file/extend_membership_successpage.html
LenaWil: Also I am curious, did you intend to use the “&” instead of "? (I am a fan, just wondering.)

Kat-zorn: I always rotate my quotation marks properly, unless it's in contractions, like the ‘it's’ above (because it's hard to teach my autocorrect to do that).

LenaWil (approved):

